PR #14, opened 2017-03-06
Added support for covers

feedfacecoffee:
Created an implementation for covers.


mnoorenberghe:
Apologies but given the current state, we don't plan to support the native UI in the near future so I don't think this PR will be necessary anymore. The good news is that development on the app has resumed though! Thanks for your patch

PR #53, opened 2017-10-22
Update Realm to 2.8.3

mnoorenberghe:
This is the earliest version that builds with XCode 9 and doesn't have the breaking changes of v3.x


robbiet480:
@mnoorenberghe Are you available on Discord?

mnoorenberghe:
> Are you available on Discord?

I am now with the same username… I was avoiding creating yet another account. I was also wondering if we should have a devs_ios channel

robbiet480:
@mnoorenberghe Great, just sent you a message

PR #59, opened 2017-10-23
Default camera category attachments to jpeg content-type

### LINKED ISSUE #39: Thumbnail not showing in notifications
Using the camera category or a url image attachment, the notification never contains the thumbnail.  I've tried setting the hide thumbnail attribute to true and false and it still doesn't work.  I did find this post (https://stackoverflow.com/a/39356505) but i'm not sure if the current codebase is using Notification Content Extension or  Notification Service Extension.

EDIT:
Looking through the code I see we are implementing both the Notification Service Extension for the notification and then Notification Content Extension for the playback of the live feed.  It does look like here https://github.com/home-assistant/home-assistant-iOS/blob/master/APNSAttachmentService/NotificationService.swift#L109 that you are setting the fileidentifier to the the attachmentURL.lastPathComponent but in documentation[0] for that method it says this doesn't work for urls, so maybe that is the issue.  It also looks like the identifier is not then being used when calling the init function https://github.com/home-assistant/home-assistant-iOS/blob/master/APNSAttachmentService/NotificationService.swift#L147.

I may be way off on all this, haven't developed in swift for iOS, this is just my poking around at the logic.

[0] - https://developer.apple.com/documentation/foundation/nsstring/1416528-lastpathcomponent

mnoorenberghe:
Mostly fixes #39 

In the case where we're a camera notification and there is no custom attachment URL, we know that we have a URL like

> \(baseURL)/api/camera_proxy/\(entityId)?api_password=\(apiPassword)

which doesn't have a suffix for UNNotificationAttachment to auto-detect the content-type.

In this case we should default to "jpeg" since that's what "camera_proxy" seems to normally return. This saves users from having to manually specify the content-type for camera stream thumbnails in most cases.

### DIFF
--- a/APNSAttachmentService/NotificationService.swift
+++ b/APNSAttachmentService/NotificationService.swift
@@ -51,6 +51,9 @@ final class NotificationService: UNNotificationServiceExtension {
                 return failEarly()
             }
             incomingAttachment["url"] = "\(baseURL)/api/camera_proxy/\(entityId)?api_password=\(apiPassword)"
+            if incomingAttachment["content-type"] == nil {
+                incomingAttachment["content-type"] = "jpeg"
+            }
         } else {
             // Check if we still have an empty dictionary
             if incomingAttachment.isEmpty {


robbiet480:
This won't cause issues with the mjpeg stream right?

mnoorenberghe:
> This won't cause issues with the mjpeg stream right?

It doesn't look like it if I follow the code but I'm not Apple to test it easily since I can't use the simulator and need to setup all the certificate stuff to test on hardware. If you're able to test it or make me a build I can test.

At some point I'd like to make it easy to show local notifications for testing in the simulator (and UI tests).

PR #61, opened 2017-10-25
Don't hardcode AppIdentifierPrefix in Keychain constructor

mnoorenberghe:
This makes it so device builds can work if the developer isn't in Robbie's team (UTQFCBPQRF).

According to https://developer.apple.com/library/content/documentation/Security/Conceptual/keychainServConcepts/02concepts/concepts.html#//apple_ref/doc/uid/TP30000897-CH204-SW1 defining `keychain-access-groups` in the entitlements means "apps would add new keychain items to that access group by default". We already have "$(AppIdentifierPrefix)io.robbie.HomeAssistant" in all of the entitlements files so it seems like specifying the accessGroup isn't necessary.

@robbiet480 Can you test that your saved login is still accessible if you upgrade to a build with this PR? According to the docs it should work and this will make it easier for other to contribute to the app.

### DIFF
--- a/APNSAttachmentService/NotificationService.swift
+++ b/APNSAttachmentService/NotificationService.swift
@@ -28,7 +28,7 @@ final class NotificationService: UNNotificationServiceExtension {
             contentHandler(request.content)
         }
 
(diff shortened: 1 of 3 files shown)


robbiet480:
LGTM!

PR #77, opened 2017-11-28
Disable webview bouncing

bjornorri:
When the webview's scrollview is allowed to bounce it sometimes scrolls
to its bouncing margins when the user intended to scroll the page loaded
in the web view. This results in ugly and annoying effects such as the
top bar of the web interface scrolling down from the status bar. Since
the loaded page has a bouncing effect anyway it seems natural to disable
this capability.


homeassistant:
Hi @bjornorri, 

It seems you haven't yet signed a CLA. Please do so [here](https://home-assistant.io/developers/cla_sign_start/?pr=home-assistant&#x2F;home-assistant-iOS#77).

Thanks!

PR #89, opened 2018-03-09
Remove invalid service data for device_tracker.see

balloob:
So first off, this is completely untested.

Starting Home Assistant 0.65, we've added service data validation to the device tracker see service. Once released, we realized that iOS app sends invalid data. This will remove those keys.

I saw one more reference to the device tracker see service, but it seems that any other app can define the data that is being sent (is that secure?)

https://github.com/home-assistant/home-assistant-iOS/blob/545ed6a7d48b70f9cb7de4c889dddfe65f48a486/HomeAssistant/AppDelegate.swift#L231

### DIFF
--- a/HomeAssistant/HAAPI.swift
+++ b/HomeAssistant/HAAPI.swift
@@ -208,10 +208,8 @@ public class HomeAssistantAPI {
 
         let locationUpdate: [String: Any] = [
             "battery": Int(UIDevice.current.batteryLevel*100),
-            "battery_status": batteryState,
             "gps": [coordinates.latitude, coordinates.longitude],
             "gps_accuracy": accuracy,
-            "hostname": hostname,
             "dev_id": deviceID
         ]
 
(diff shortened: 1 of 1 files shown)


balloob:
Would love to get this released soon.

Added a temp workaround to Home Assistant to not break all iOS apps: https://github.com/home-assistant/home-assistant/pull/13020

mnoorenberghe:
I also didn't test this but it looks really straightforward. I added a commit to clean up the now unused variables too.

I tracked this code down and AFAICT these values were unused by the server and it relies on the battery state sent with the `identify` request. The hostname doesn't seem like it was used at all.

Note that while I have commit access, I don't have access to do releases so @robbiet480 would still have to do that.

robbiet480:
About to release this. I re-added hostname as `host_name` as the device tracker code does use it to decide the friendly name of the device in addition to the device ID.

balloob:
Robbie, can we give @mnoorenberghe access to do releases if he wants?

PR #120, opened 2018-08-19
Support the new auth system.

blackgold9:
This PR adds support for authenticating with Hass's new auth system. It begins moving some settings out of the HomeASsistantAPI space and into A dedicated settings store, and sets up classes to managed the authentication flow, and to deal with token management. 

I Also started refactoring some of our longer methods into smaller chunks, and removed the concept of a sharedInstance of the API. You can now grab the authenticated API (if available) via HomeAssistantAPI.authenticatedAPI() or authenticatedAPIPromise (not sold on the name i used).

Token management on requests is done via a token manager that takes over as a decorator of all requests, and as the object determining retry logic.

As it stands the user has to log in twice, as there's no concept of authenticating the webview by token yet.


HomeAssistant/AppDelegate.swift
```diff
@@ -25,12 +26,13 @@ let prefs = UserDefaults(suiteName: Constants.AppGroupID)!
...
     var window: UIWindow?
-
+    var safariVC: SFSafariViewController?
     func application(_ application: UIApplication,
                      didFinishLaunchingWithOptions launchOptions: [UIApplicationLaunchOptionsKey: Any]?) -> Bool {
         Fabric.with([Crashlytics.self])
-        HomeAssistantAPI.sharedInstance.Setup(baseURLString: keychain["baseURL"], password: keychain["apiPassword"],
-                                              deviceID: keychain["deviceID"])
+
+//        HomeAssistantAPI.sharedInstance.setup(baseURLString: keychain["baseURL"], password: keychain["apiPassword"],
```
robbiet480: Plz remove any stale lines like this

HomeAssistant/AppDelegate.swift
```diff
@@ -60,6 +62,10 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
         return true
     }
 
+    func ensureToken() {
```
robbiet480: This is an empty function?

HomeAssistant/Authentication/AuthenticationController.swift
```diff
@@ -0,0 +1,97 @@
...
+//  HomeAssistant
+//
+//  Created by Stephan Vanterpool on 8/11/18.
+//  Copyright © 2018 Robbie Trencheny. All rights reserved.
+//
+
+import Foundation
+import PromiseKit
+import SafariServices
+let kAuthenticationPath = "/auth/authorize?response_type=code&client_id=https%3A//blackgold9.github.io"
```
robbiet480: Let's change this `client_id` to be like` `https://home-assistant.io/ios`?

HomeAssistant/Authentication/AuthenticationController.swift
```diff
@@ -0,0 +1,97 @@
...
+            let responseTypeQuery = URLQueryItem(name: "response_type", value: "code")
+            let clientIDQuery = URLQueryItem(name: "client_id", value: "https://blackgold9.github.io/home-assistant-iOS")
+            let redirectQuery = URLQueryItem(name: "redirect_uri", value: "homeassistant://auth-callback")
+            components?.queryItems = [responseTypeQuery, clientIDQuery, redirectQuery]
+            if let newURL = try components?.asURL() {
+                let safariVC = SFSafariViewController(url: newURL)
+                if #available(iOS 11.0, *) {
+                    safariVC.dismissButtonStyle = .cancel
+                } else {
+                    // Fallback on earlier versions
```
robbiet480, line 46: So what about iOS 10?

HomeAssistant/Classes/Components/Climate.swift
```diff
@@ -56,7 +56,7 @@ class Climate: Entity {
     }
 
     func TurnFanOn() {
-        _ = HomeAssistantAPI.sharedInstance.CallService(domain: "climate",
+        _ = HomeAssistantAPI.authenticatedAPI()?.CallService(domain: "climate",
```
robbiet480: What happens if authenticatedAPI returns nil?

HomeAssistant/HAAPI.swift
```diff
@@ -216,6 +210,65 @@ public class HomeAssistantAPI {
         }
     }
 
+    public enum HomeAssistantAPIError: Error {
+        case notAuthenticated
+    }
+//    static var internalAuthenticatedAPI: Promise<HomeAssistantAPI>?
```
robbiet480: Remove stale code

HomeAssistant/Authentication/AuthenticationController.swift
blackgold9: It says done. Figured that was decent enough of a graceful fallback.

HomeAssistant/AppDelegate.swift
blackgold9: yah. Srry. Taking a clean up pass through it again.

HomeAssistant/Authentication/AuthenticationController.swift
blackgold9: We can change it, but that address needs to host a page with a specific tag in it's header. Don't know how to make the current home assistant webpage do that.

HomeAssistant/Classes/Components/Climate.swift
blackgold9: fails silently. I think that's ok as other parts of the app should handle re-directing the user to authenticate if we are in that state.

robbiet480:
Thanks! 🐬 🍪 💯

blackgold9:
🎉

PR #133, opened 2018-08-31
Reload webview to active URL if it changed when in the background

blackgold9:


blackgold9:
Hangon on merging this. I Think it's making it impossible to enter 2fa codes.

blackgold9:
Ok, it's fixed now.

PR #142, opened 2018-09-07
Added working token transfer!

blackgold9:
This works end to end, or at least, it will after https://github.com/home-assistant/home-assistant-polymer/pull/1644 is merged.


blackgold9:
Don't merge until  home-assistant/home-assistant-polymer#1644 is merged.

Issue #29, opened 2017-04-26
Leaving known zone doesn't seem to put you into "not_home" state


cmsimike:
Hi,

I have zone notifications on so I can debug when the app thinks I enter and leave a hass zone. This morning I noticed it said I left my home zone, and 6 minutes(ish) later it said I entered my work zone. Looking at the ui however doesn't indicate my device state changed to 'not_home' during that 6 minutes.

cmsimike:
I noticed it put me into 'not_home' this morning so I guess it just isn't consistent?

bagobones:
6 min commute to work? I wonder if it isn't just a case of background updates being very slow... Remember it only updates on significant location changes when in the background.

cmsimike:
@bagobones i mean if the app can figure out if it left the zone, doesn't that imply the significant change happened? and it isn't a 6 minute commute - about 20 mins total, so it took some time for it to figure out i was out of my home zone but 6 mins after that it knew it was at work.

bagobones:
Testing today I do see it update my location once on the way to work so I am not_home however the update is somewhere in-between work and home.. I have been at the office for over an hour and another update has not been sent apparently.


I am going to say this issue is that Background update AND GEOfencing are not working correctly.

lucasweb78:
I'm seeing this issue as well. When I leave my home zone in the morning I get the `exited home` notification but do not see the location change pushed to hass for a considerable period. If I force a location change update then hass does update.

I do not see this issue when I enter my home zone, both the notification and hass update happen almost immediately and at the same time.

gieljnssns:
I have the same problem...

robbiet480:
Can everyone here please retest with 1.5.0 and report back? Thanks!

ukgrant:
I was also experiencing this issue. After upgrading to 1.5.0 exiting zones doesn't seem to be firing at all.
Entering a zone works fine, but exiting a zone always produces "Not updating zone... because the database says nothing has changed" in the log.

App version 1.5.0
iOS 12
iPhone X
Update sources: Zone enter/exit only (all others disabled)

ukgrant:
I installed 15.0 (17) this morning. Exiting a zone is now triggering. However "leaving home" automation's are not triggered.

When crossing the geo-fence boundary, the app sends a 'leave zone' event to home assistant with the coordinates of the centre of the zone. Home assistant shows the device as 'away', but inside the home region on the map, which is keeping the lights on.

[...]

PR #171, opened 2018-11-23
Set location name to not_home on zone exit for all zones

ukgrant:
Addresses issue #29 where region monitoring does not send the not_home state when leaving a zone. Previously this only happened when leaving the "home" zone. This commit adds this state for any zone, so that region exit automations can be triggered in a timely manner.

### DIFF
--- a/Shared/API/Models/DeviceTrackerSee.swift
+++ b/Shared/API/Models/DeviceTrackerSee.swift
@@ -65,8 +65,19 @@ public class DeviceTrackerSee: Mappable {
                 self.LocationName = LocationNames.Home
             case .RegionExit, .GPSRegionExit:
                 self.LocationName =  LocationNames.NotHome
+                self.ClearLocation()
             case .BeaconRegionExit:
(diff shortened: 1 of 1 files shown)


homeassistant:
Hi @ukgrant, 

It seems you haven't yet signed a CLA. Please do so [here](https://home-assistant.io/developers/cla_sign_start/?pr=home-assistant&#x2F;home-assistant-iOS#171).

Thanks!

blackgold9:
Thanks for this. we really appreciate people helping out. That said, this isn't quite the right approach. home/nothome is specifically for the home zone as i understand it after looking at the device_tracker code. I've just added code to send the gps location on region exit instead of the zone's gps location. I think this should fix the problem you're trying to solve. Can you check it out?

ukgrant:
@blackgold9, no worries. Thanks for looking into this. I agree that sending the GPS is a good approach. I have tested your latest change this morning and all zones exited correctly.

PR #176, opened 2018-12-07
Do nothing on non-home iBeacon exits for now

jjlawren:
Currently, non-home zone iBeacon exits will unintentionally result in a change to `home`. This skips exit logic for now until it's decided how to properly handle these triggers.


ukgrant:
I just realised that this won't "do nothing". This will still send a beacon exit event for both home and non home beacons. This just removes the 'consider home' functionality. To fully block exit events for beacons you need to add a condition in RegionManager https://github.com/home-assistant/home-assistant-iOS/blob/master/HomeAssistant/Classes/RegionManager.swift

PR #178, opened 2018-12-10
Add 'GPS Enter' event to exclude list for location coordinate reporting…

ukgrant:
Add 'GPS Enter' event to exclude list for location coordinate reporting. Enter events will send zone coordinates instead.

GPS exit events remain sending actual coordinates (outside the zone).

This is due to a user reporting their device triggering an enter event, but sending coordinates outside of their configured zone in home assistant. iOS will automatically bump any ranges up to 100m as this is the minimum.

The alternative work around is for the user to set their zone to 100m. The end result is the same, where the phone will trigger at the 100m mark, but this patch may reduce confusion.


ukgrant:
I've thought about this and it's probably best that users are required to set zone radius to 100m.

mgiako:
Please merge this, my automations do not start when I enter the home zone!

PR #180, opened 2018-12-12
Actually skip iBeacon exit events

jjlawren:
Followup on https://github.com/home-assistant/home-assistant-iOS/pull/176 which didn't actually achieve what it was intended to do.


ukgrant:
Looks good. To confirm, this will prevent all iBeacon exit events from being sent to the server.

PR #203, opened 2019-01-23
Add pod repo update step to install instructions

amelchio:
I had to do this additional step to get a working development setup.


andersonshatch:
Ran into this today too. `pod install --repo-update` is perhaps cleaner as a one-line step.

amelchio:
Thanks, I changed it.

PR #209, opened 2019-02-04
Revert #176 (partial) and #180 to restore iBeacon exit handling

jjlawren:
Knock-on effects were not considered and these should be reverted until there is more time to sufficiently test new behavior.

#176 is only partially reverted as the non-home iBeacon region handling was incorrect.


jwelter1971:
Has this made it into the 2.0 beta as it seems to still not trigger a zone exit due to ibeacons?

xppx99:
Hello!
Would at least be possible to have the option to select the behavior (either considering ibeacon zone leaves or ignoring it like it is now)?
Thanks in advance

PR #210, opened 2019-02-04
Only ask to re-enable permissions the first time after they are disabled

### LINKED ISSUE #177: "Allow Motion" Popup keeps re-appearing
Version 1.5.0 (31)
Every time I start the iOS app since the beginning of the beta I have been getting this Allow Motion popup. I do not use any location features of the Home Assistant app so I have everything disabled. But when I click deny in this popup it does not seem to save my choice after restarting the app
image

amelchio:
I think this is closer to what we discussed in Discord: permissions are now only ensured if the status has changed. So the popups should only show when permissions have just been revoked.

Fixes #177


h4nc:
Do we Need or update the iOS App for that and if yes when will we get this update?

robbiet480:
@h4nc This will be in 1.5.1 which I am planning to submit no later than Friday.

h4nc:
Thanks

melikyuksel:
Is 1.5.1 coming at some point? Two months and counting.

Issue #108, opened 2018-07-11
Signing issues when building locally


torarnv:
I'd like to work on the iOS app, build it, and test it on my iPhone, but it seems like a lot of the signing stuff in the Xcode project is hard-coded to a certain development team and certain provisioning profiles. How are contributors supposed to build the app? Should I register provisioning profiles with my own app id and app group id, etc? Or is there a way to build the app with the existing profiles/app group settings?

@blackgold9 how are you solving this?

torarnv:
Perhaps it would be possible to define the team and provisioning profiles in an `xcconfig` file, with comments on what profiles/enablements need to be registered, so that it's easy to hot-swap for another developer for on-device development

blackgold9:
PRs welcome :)

torarnv:
:smile:

robbiet480:
I've said this in #devs_ios previously, but the primary issue here is that I have an individual account which means that I can only have my certificates which is why everything is hardcoded. You can go ahead and replace the certificate settings on your local version to build and test but please try to reset the project file containing the settings to defaults when submitting PRs. Hopefully someday this app will be under the HA organization and we can have multiple team members.

torarnv:
The app is now under the HA organization. Does that mean the app can now be built without replacing certs?

robbiet480:
What organization? It’s always been under the HA GitHub organization & the app is still published under my Apple Developer account.

torarnv:
Ah, my bad, I was looking at this and thought the GitHub org was the one that needed changing. I just pulled the latest master and can confirm what you're saying. I'll look into moving some of that into an xcconfig file to at least make it easier to customize. Thanks!

robbiet480:
One thing to note if you or anyone gets to this someday: The app now features the special critical alerts entitlement which must be requested from Apple. You _should_ be able to just remove it from the entitlements file and be on your way, but it may also require removing the request for critical alerts in the code too.

robbiet480:
Fixed via #386

PR #386, opened 2019-05-28
Move configuration from Xcode project to .xcconfig files

torarnv:
This patch set moves key configuration items from Xcode project to .xcconfig files, so that they can be easily overridden. This is useful for building the App with a different signing and provisioning setup than the default one used for the App Store, fixing #108.

For example, using an override file like this allows building for device:

```
DEVELOPMENT_TEAM = ABCDEFGH123
BUNDLE_ID_PREFIX = com.foo

CODE_SIGN_STYLE = Automatic
PROVISIONING_PROFILE_SPECIFIER =
```

#### Some remaining issues

- [ ] Remove leftover occurences of `io.robbie` where possible and replace with `BUNDLE_ID_PREFIX`
```
HomeAssistant/Resources/GoogleService-Info-Beta.plist:  <string>io.robbie.HomeAssistant.beta</string>
HomeAssistant/Resources/GoogleService-Info-Development.plist:   <string>io.robbie.HomeAssistant.dev</string>
HomeAssistant/Resources/GoogleService-Info-Release.plist:   <string>io.robbie.HomeAssistant</string>
HomeAssistant/Resources/gallery.ckcomplication/complicationManifest.json:  "client ID" : "io.robbie.HomeAssistant.dev.watchkitapp.watchkitextension"
HomeAssistant/Views/WebViewController.swift:                self.userActivity = NSUserActivity(activityType: "io.robbie.HomeAssistant.frontend")
Shared/API/HAAPI.swift:                seal.reject(NSError(domain: "io.robbie.HomeAssistant", code: 500, userInfo: nil))
```
There are also a bunch of instances under `fastlane/`, but since those are only used by @robbiet480 it's probably not any point in replacing those, plus I'm not sure if it's even possible.    

---
- [ ] Deal with `com.apple.developer.usernotifications.critical-alerts` entitlement. Some options:
 - Use [string instead of bool](https://gist.github.com/torarnv/38a5969e9a860e4d3ca4cc004872077d). Pleases Xcode, but might break app store review.
 - Make the plist _key_ empty to disable the entitlement. Pleases Xcode, but might break app store review.
 - [Preprocess](https://stackoverflow.com/a/39866059/2761869) the plist file
 - [Postprocess](https://stackoverflow.com/a/32704702/2761869) the plist file

---
- [ ] Remove empty `PROVISIONING_PROFILE_SPECIFIER` entries from project file

The targets that have `PROVISIONING_PROFILE_SPECIFIER = ""` get `Automatic` provisioning right now. If we remove the entry they would fall back to the `PROVISIONING_PROFILE_SPECIFIER = ${PROVISIONING_PROFILE_SPECIFIER_PREFIX} ${PRODUCT_BUNDLE_IDENTIFIER}
` in the configuration file. Not sure if those targets support that (that there's an explicit manually created provisioning profile for them).

---
- [ ] Consider using automatic signing for all targets

This would reduce the override file to simply:

```
DEVELOPMENT_TEAM = ABCDEFGH123
BUNDLE_ID_PREFIX = com.foo
```


robbiet480:
Okay, initial thoughts on the remaining TODOs:

- Not going to change any of the `GoogleService-Info-*.plist` files, since without those push notifications won't work unless a developer duplicates the entire Firebase setup including cloud function.
- Don't know of any dynamic way to change `HomeAssistant/Resources/gallery.ckcomplication/complicationManifest.json` but open to ideas!
- Once this is merged, we can get fastlane to use the xcconfig files too. I'll only change the build lanes though, the `certs` and `push` lanes will stay the same
- I think leaving `PROVISIONING_PROFILE_SPECIFIER` as is will be fine since the only targets it is empty in are non building/deployment targets such as the tests & Shared ([screenshot](https://cdn.discordapp.com/attachments/582957157545148427/583050621423321141/unknown.png) of current config)
- I don't want to use Automatic signing for any deployable targets because I use fastlane's codesigning tools to allow me to store a backup copy of all the iOS certificates in a private GitHub repo that key people have access to so that in case the [bus factor](https://en.wikipedia.org/wiki/Bus_factor) kicks in those key people can continue releasing updates without needing me (they also have access to App Store Connect).

robbiet480:
Will deal with remaining issues in beta branch

PR #388, opened 2019-05-29
Make critical alerts optional

torarnv:


torarnv:
The code that uses the critical alert probably needs to check the entitlement first. or if swift has idefs, we can pass -D ENABLE_CRITICAL_ALERTS=${ENABLE_CRITICAL_ALERTS} when building to remove the code at build time.

PR #541, opened 2020-02-06
Change mdi icons for floors ascended and descended

gerard33:
There are new stairs up and down icons available and these are part of HA 0.105
https://materialdesignicons.com/icon/stairs-up
https://materialdesignicons.com/icon/stairs-down

### DIFF
--- a/Shared/API/Webhook/WebhookSensors.swift
+++ b/Shared/API/Webhook/WebhookSensors.swift
@@ -237,7 +237,7 @@ public class WebhookSensors {
             return nil
         }
 
-        return WebhookSensor(name: "Floors Ascended", uniqueID: "pedometer_floors_ascended", icon: "mdi:slope-uphill",
+        return WebhookSensor(name: "Floors Ascended", uniqueID: "pedometer_floors_ascended", icon: "mdi:stairs-up",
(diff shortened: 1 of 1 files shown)


TomBrien:
This will be a breaking change for anyone who's phone automatically updates but isn't running >0.105.

gerard33:
That's true, but only when re-installing the mobile app integration isn't it? Otherwise the existing entities and mdi icons will not be updated. And I guess using the latest version of HA is always recommended?

TomBrien:
Hopefully @robbiet480 can confirm but I’m pretty sure the icon is set which each sensor update. See the similar code for battery icons and BSSID

robbiet480:
Yeah I don't really like this since it could cause icons to go missing if people are on older versions of HA. Not opposed to merging this at some point, but let's just wait until 0.105 has been out in the world for more than a few weeks.

robbiet480:
It’s been long enough I feel comfortable merging this, but to be safe can you modify this a bit to set the icon based on the HA version? If you aren’t a Swift dev I can do it, just let me know!

gerard33:
Not a Swift dev so would be great if you can add that check.

robbiet480:
Closing in favor of #678

PR #587, opened 2020-05-12
feat: clear_notification

boesing:
As already requested in some issues, I'd like to try to implement this feature.

I am not sure if that works. I am neither a swift developer nor can I make XCode run properly on my MacBook due to annoying issues with `Eureka` (getting a segmentation fault... while compiling).

I've had a bit of research and I think that this might work as expected.
I am unsure about the usage of `userInfo["apns-collapse-id"]`, so feedback is very welcome.

Resources:
https://medium.com/@sebastianosiski/implementing-removable-remote-notifications-on-ios-a17d74832bde


homeassistant:
Hi @boesing, 

It seems you haven't yet signed a CLA. Please do so [here](https://home-assistant.io/developers/cla_sign_start/?pr=home-assistant&#x2F;iOS#587).

Thanks!

boesing:
@robbiet480 Would love to get some feedback here.
Is there anything missing to make this PR find its way into one of the next releases?

robbiet480:
Yeah this isn't gonna work. `apns-collapse-id` is implemented in Apple`s Push Notification Service (APNS) and the value isn't passed down to us. Not sure why collapse doesn't work as is.

boesing:
Collapse itself works, the message is replaced in current Versions. But the message says "clear_notification" as the command is not implemented 😅

Well, so there is no way to remove a notification for iOS? Only replacing so far?

robbiet480:
That's correct. Docs should be updated to clarify that iOS doesn't yet support `clear_notification`.

skrenes:
@robbiet480 do you mean the iOS app doesn't yet support this or that you don't think iOS itself supports this, because I found [this reference](https://stackoverflow.com/a/55646607) that suggest that it can be done by using the same `apns-id` in the request that was received in the initial response.

robbiet480:
@skrenes I don’t think that will work for us because Firebase doesn’t return the APNS ID to us afaik. Only their ID.

boesing:
Hm, it seems that its really impossible to receive the `apns-id` from firebase.
But what about that approach I've seen in that medium post I've mentioned above and (quite the same logic in another [medium post](https://medium.com/the-guardian-mobile-innovation-lab/how-to-replace-the-content-of-an-ios-notification-2d8d93766446))?

Receiving all notifications which are still displayed in the notification center with `UNNotificationCenter::getDeliveredNotifications(completionHandler:)` and iterating over all notifications to search for a notification with a specific collapse ID (which we know if we sent it with the initial notification)?
Could use `UNNotificationCenter::removeDeliveredNotifications(withIdentifiers:)` to remove that notification we found.

OnFreund:
Here's an interesting solution that passed the `apns-collapse-id` in a separate parameter so that the app gets is:
https://medium.com/@sebastianosiski/implementing-removable-remote-notifications-on-ios-a17d74832bde

boesing:
@OnFreund please read the other comments here. Thats already known but that ID has to be received from firebase which is not possible. Its not a Static ID like the collapse ID. 
I've thought the same...

robbiet480:
Going to close this for now until we come up with a better solution. Not opposed to what I’ve heard thus far, just needs to actually work and not be too hacky.

boesing:
Thx for having a look on this.

PR #602, opened 2020-06-07
Fix building MaterialDesignIcons

zacwest:
This fixes executing `Tools/BuildMaterialDesignIconsFont.sh` at least on 10.15.5 (Catalina). There are a few underlying issues here:

1. The `fontname.py` dependency was updated to require python3+, so we need to both install python3 variants and execute using the python3 binary.
2. The `MaterialDesign-Webfont` dependency, for versions ≥5.0, no longer parses successfully with the version of SwiftGen in Iconic.

```bash
$ Vendor/SwiftGen/build/swiftgen/bin/swiftgen icons /…/materialdesignicons-webfont-v5.0.45.ttf
// Generated using SwiftGen, by O.Halligon — https://github.com/AliSoftware/SwiftGen

import UIKit

// No icon found
```

I started down the rabbit hole for 2 but couldn't find an obvious path forward -- it seems like SwiftGen has dropped icon-font building support, so updating to the latest version there doesn't seem fruitful.

While I was in here, I made the script a little less sensitive to executing location by wrapping it in a `pushd`/`popd` block.

### DIFF
--- a/Tools/BuildMaterialDesignIconsFont.sh
+++ b/Tools/BuildMaterialDesignIconsFont.sh
@@ -3,11 +3,11 @@
 # Just run this script and it should handle everything.
 # You do need enough of a Python environment setup to be able to pip install things.
 
-cd Tools
+pushd `dirname $0`
 
 # Ensure fonttools is installed
 echo "Ensuring fonttools is installed via pip"
-pip install fonttools
+pip3 install --user fonttools
 
 if [ ! -f fontname.py ]; then
   # Download the fontname script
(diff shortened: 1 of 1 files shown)


homeassistant:
Hi @zacwest, 

It seems you haven't yet signed a CLA. Please do so [here](https://home-assistant.io/developers/cla_sign_start/?pr=home-assistant&#x2F;iOS#602).

Thanks!

robbiet480:
@zacwest FYI our icon library, Iconic, uses a old and forked version of SwiftGen to do icon font generation. I've had a PR for a while now that adds it to SwiftGen proper at https://github.com/SwiftGen/SwiftGen/pull/638.

zacwest:
Oh thank goodness. The current one in Iconic is like Swift 2 so I was worried about that!

PR #604, opened 2020-06-07
Action Configurator: Example Trigger Preview

zacwest:
When editing an Action, the configurator currently shows a preview of what the button will look like, but it doesn't give any direct indication what the _event_ it fires will look like. One must look to the documentation to get an idea how to structure the automation.

This adds a preview of the yaml one would need to write as an automation, using the same flow as sending it to the server to avoid it getting stale over time, and allows the user to share it out via copying, email, etc.

This pulls together the different 'previews' in the configurator to be executed together:
- the title of the view controller (which wasn't updated live before)
- the actual preview button the user can tap
- this new yaml trigger preview

This also removes the toolbar with the help button from the action edit screen. It didn't work, it takes up a lot of vertical space where it's at a premium with the keyboard open, and the previous screen already has the same link/button.

Simulator Screen Shot - iPhone 11 Pro - 2020-06-07 at 16 47 10
Simulator Screen Shot - iPhone 11 Pro - 2020-06-07 at 16 47 12

### DIFF
--- a/HomeAssistant/Resources/en.lproj/Localizable.strings
+++ b/HomeAssistant/Resources/en.lproj/Localizable.strings
@@ -445,6 +445,8 @@
 "actions_configurator.rows.text.title" = "Text";
 "actions_configurator.rows.text_color.title" = "Text Color";
 "actions_configurator.title" = "New Action";
+"actions_configurator.trigger_example.title" = "Example Trigger";
+"actions_configurator.trigger_example.share" = "Share Contents";
 "alerts.alert.ok" = "OK";
 "alerts.confirm.cancel" = "Cancel";
 "alerts.confirm.ok" = "OK";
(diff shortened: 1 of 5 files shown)


HomeAssistant/Resources/en.lproj/Localizable.strings
```diff
@@ -445,6 +445,8 @@
 "actions_configurator.rows.text.title" = "Text";
 "actions_configurator.rows.text_color.title" = "Text Color";
 "actions_configurator.title" = "New Action";
+"actions_configurator.trigger_example.title" = "Example Trigger";
+"actions_configurator.trigger_example.share" = "Share Contents";
```
zacwest, line 449: I wordsmithed this a few times and didn't land on anything much better than "Share Contents".

PR #621, opened 2020-06-10
Ensure that OneTimeLocationManager doesn't use SignificantLocationChange

### LINKED ISSUE #589: Issue with location accuracy after 13.4 update

Model Name: iPhone 11 Pro
Software Version: 13.5 
App version: 2019.1

0.110.1 w/temporary webhook.py fix for sim2 errors 

- https://github.com/brg468/core/tree/mobile_app_webhook_fix/homeassistant/components/mobile_app
- https://github.com/home-assistant/core/issues/35812#issuecomment-631737472

**Describe the bug**
I am seeing my accuracy flail all over the place. Sometimes it goes as high as 2500, then drops down to 65. I can't seem to get my accuracy under 65, although my wife's iPhone X has no problem getting down to 10. 

My update type will change to Significant Location Change even though I haven't moved because it doesn't think I'm 2 miles away any more. 

Others on the HA forum are experiencing similar issues - https://community.home-assistant.io/t/gps-accuracy-after-upgrade-to-ios-13-4/181585/5

**To Reproduce**
Install app, allow initial sensors to be created, review GPS accuracy. 

**Expected behavior**
Stable location

**Screenshots**
The blue larger circle is my iPhone 11 Pro and the smaller green circle is an X. Both are on the same WiFi and cellular carrier. 
image

TomBrien:
We have seen multiple people report poor accuracy with the app including multiple reports of background fetch location accuracy being much worse than 100 m (2500 m has been reported). Background fetch, along with everything except zones, iBeacons, and significant location update, uses `OneShotLocationManager` which should never return a location with accuracy of worse than 200 m and aspires to achieve better than 100 m. 

I believe this is because `CoreLocation` works app wide and setting one manager to `startMonitoringSignificantLocationChanges()` causes all managers to disregard `desiredAccuracy` and `distanceFilter` as explained [here](https://developer.apple.com/documentation/corelocation/getting_the_user_s_location/using_the_significant-change_location_service).

Attempt to address this by calling `stopMonitoringSignificantLocationChanges` with the `OneShotLocationManager` before doing the one-shot update and then turning monitoring back on at all exit conditions. 

Potentially fixes #589

### DIFF
--- a/HomeAssistant/Classes/OneShotLocationManager.swift
+++ b/HomeAssistant/Classes/OneShotLocationManager.swift
@@ -19,6 +19,7 @@ public class OneShotLocationManager: NSObject, CLLocationManagerDelegate {
     public init(onLocation: @escaping OnLocationUpdated) {
         onLocationUpdated = onLocation
         super.init()
+        locationManager.stopMonitoringSignificantLocationChanges()
         locationManager.desiredAccuracy = kCLLocationAccuracyHundredMeters
         locationManager.distanceFilter = kCLLocationAccuracyHundredMeters
         locationManager.delegate = self
@@ -35,6 +36,7 @@ public class OneShotLocationManager: NSObject, CLLocationManagerDelegate {
 
             Current.Log.verbose("LocationManager: Got location stored on manager \(location)")
             onLocationUpdated(location, nil)
+            locationManager.startMonitoringSignificantLocationChanges()
             locationManager.delegate = nil
         }
     }
@@ -42,6 +44,7 @@ public class OneShotLocationManager: NSObject, CLLocationManagerDelegate {
     public func locationManager(_ manager: CLLocationManager, didUpdateLocations locations: [CLLocation]) {
         Current.Log.verbose("LocationManager: Got \(locations.count) locations, stopping updates!")
         onLocationUpdated(locations.first, nil)
+        locationManager.startMonitoringSignificantLocationChanges()
         manager.delegate = nil
     }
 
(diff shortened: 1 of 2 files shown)


zacwest:
Apple's documentation says:

> When requesting high-accuracy location data, the initial event delivered by the location service may not have the accuracy you requested. The location service delivers the initial event as quickly as possible. It then continues to determine the location with the accuracy you requested and delivers additional events, as necessary, when that data is available.

I think what's happening is that (if significant location change is a factor here), Core Location has a low-accuracy, cached value that the app immediately discards (when reading the cached value) which CL then sends on the delegate method.

A good fix here would be to race:
1. A good-accuracy location
1. Timing out and using the best location we've got in e.g. 5-10 seconds
and whichever wins is the location that's fired. Since this is a background event, we've got 30-ish seconds depending on Apple's mood to get better location data, and we want to leave a buffer for networking setup and such.

I think this would also be a great opportunity to use a background session to make sure the network event survives the background timer.

TomBrien:
Agreed that this isn’t the most rugged way forward. New PR looks nice. I was working on what was reported in a vaguely similar case [here](https://stackoverflow.com/questions/26839170/two-cllocationmanager-interfere). I must say I never see the reported bug. It does seem to be an iPhone 11(pro) thing. The 200 m cut works well on my phone

zacwest:
I would say, given how wonky Core Location is, there's still a significant chance that they impact each other; hopefully we can get work around any sig-monitoring -> 2500m issues by giving CL some time to spin up a tighter radius.

TomBrien:
Agreed. In the old one shot I toyed with aldehyde disregarding the first (regardless of cache) and using the last in value not first in.

PR #623, opened 2020-06-11
New One Shot Location Logic

zacwest:
- Adds a thin wrapper around the old logic, which allows users to swap between the two, reducing the scariness of this PR.
- Wraps manual location updating in the same background task logic as background ones.
- Adds background time remaining (when available) as a part of the background task logic, so we can pipe that into the location updating logic, too.

This new logic is focused around: how many updates can we get and how much accuracy can we hit if we give Location Services a few more seconds to feed us location data? I've plugged in some random guesses at how long we're willing to tolerate these things.

Indications are that iOS is willing to toss us things like 2500m accuracy in some situations, which is a heck of a lot of drift for a background fetch update. We're being given an opportunity to do some background work, so it feels like a good time to try and use it, too.

It might be worth considering adding some analytics here to try and get a better handle on how often we have either no location, a mediocre location, or a perfect location.


robbiet480:
Analytics could be easily added via Firebase.

PR #625, opened 2020-06-11
Replace hassio.local placeholders

TomBrien:
Update the placeholder for the manual onboarding box with `homeassistant.local:8123`. I believe all other instances are covering on the fly by localise? Also did a comment because why not


robbiet480:
@TomBrien This also needs to be updated on Lokalise.

PR #628, opened 2020-06-12
Page Zoom setting

zacwest:
Scales the contents of the web view to a multiple of normal. This behaves similar to the +/- button in Safari.

This is different than both:
- `document.body.style.zoom` (which causes the header to be positioned wrong)
- `-webkit-text-size-adjust` (which causes items to be badly centered)

The setting looks like:

IMG_6596

The difference between e.g. 150%, 100% and 85% looks like:

IMG_6599


TomBrien:
Nice

PR #630, opened 2020-06-12
Make Reset also trigger onboarding

zacwest:
This now attempts to revoke the token, with a bail-out mechanism if it's taking too long, and triggers onboarding once everything is reset.


TomBrien:
Very here for this one

Issue #554, opened 2020-03-01
Request: Use consistent user agent for all requests


efirestone:
I'd like to reverse proxy HA requests, but this is harder than it should be as there is no support for paths (subfolders) in the `base_url`, and apparently [never will be](https://github.com/home-assistant/architecture/issues/156#issuecomment-478183627).

For me, supporting proxying for just the iOS app would be sufficient, and I believe that can be achieved by proxying based on the user agent of the request. The issue right now is that many requests come through with a custom agent from the app, like so:

```
Home%20Assistant/69 CFNetwork/1121.2.2 Darwin/19.3.0
```

But many others come through with a more generic one:

```
Mozilla/5.0 (iPhone; CPU iPhone OS 13_3_1 like Mac OS X) AppleWebKit/605.1.15 (KHTML, like Gecko) Mobile/15E148
```

Presumably the latter is coming from requests that are initiated by a WKWebView instance directly (and the types of requests, like images and other resources) would support this.

It looks like we can use the custom user agent for these requests as well by setting it via [`WKWebView.customUserAgent`](https://developer.apple.com/documentation/webkit/wkwebview/1414950-customuseragent).

Thanks for your consideration!

efirestone:
I can't seem to get the dev app to launch in the sim for some reason, but I believe this might be the entire change:
```
 import UIKit
+import Alamofire
 import WebKit
 import KeychainAccess
 import PromiseKit
@@ -117,6 +118,7 @@ class WebViewController: UIViewController, WKNavigationDelegate, WKUIDelegate {
         config.userContentController = userContentController
 
         self.webView = WKWebView(frame: self.view!.frame, configuration: config)
+        self.webView.customUserAgent = SessionManager.defaultHTTPHeaders["User-Agent"]
         self.webView.isOpaque = false
         self.view!.addSubview(webView)
```

robbiet480:
I guess i'm not opposed. It's not like the built in browser ever sees much more than Home Assistant UI and afaik we don't do any UA detection there. Mind submitting a PR for this so you get the credit? I'll test it locally first.

efirestone:
PR's open. Thanks!

BigCabbage:
@efirestone , can you share how you configured how you used the distinct user agent to reverse proxy HA? There are many, many users who are looking for a way to do this without using a subdomain.

If you can share an example, you or I might be able to contribute to the page which currently described how to reverse proxy using a subdomain so it will be available for everyone...

[...]

PR #632, opened 2020-06-12
Append User-Agent rather than resetting it

### LINKED ISSUE #651: New userAgent does not differentiate between iPad and iPhone
The newest update overwrites the userAgent completely, removing any relevant browser information for the device being used. This breaks any sort of filtering based on the type of device being used ie iPhone or iPad, Mobile, etc

This goes against what the userAgent is supposed/expected to contain https://developer.chrome.com/multidevice/user-agent

Model Name: iPhone 8
Software Version: 13.3.1
App version: 2020.1 (17)

0.111.1

**Describe the bug**
The new userAgent completely removes any information about the current browser. 

**To Reproduce**

**Expected behavior**
The userAgent contains all the information that actually pertains to the browser being used.

**Screenshots**

**Additional context**
This would have been fixed with PR #632 , but that was closed without a real explanation as to why

zacwest:
Refs #554, #555. Fixes #651.

### DIFF
--- a/HomeAssistant/Views/WebViewController.swift
+++ b/HomeAssistant/Views/WebViewController.swift
@@ -117,9 +117,9 @@ class WebViewController: UIViewController, WKNavigationDelegate, WKUIDelegate {
         }
 
         config.userContentController = userContentController
+        config.applicationNameForUserAgent = SessionManager.defaultHTTPHeaders["User-Agent"]
 
         self.webView = WKWebView(frame: self.view!.frame, configuration: config)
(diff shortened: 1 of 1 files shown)


robbiet480:
Gonna close this, I want to leave unique UA. There's no frontend feature detection so users only need to adapt their cards to new UA.

maykar:
Just noticed this pull request. While the frontend itself has no user agent detection: custom cards, sites in iframe cards, and hassio-addons do. I'm sure those can't be the only examples. I would hope you'd reconsider this position as basic browser info shouldn't be stripped from user agents. Thanks for considering.

PR #647, opened 2020-06-15
Handle and persist theme color

### LINKED ISSUE #557: Status bar does not change color on `frontend.set_theme` since 0.106
Since the release of `0.106` of Home Assistant, the status bar no longer changes color to match the theme when `frontend.set_theme` is called.

`0.106` locked down the set_theme call to admins, so I suspect there are some bugs related to this change (https://github.com/home-assistant/core/pull/31654).

So far, I've discovered there is a frontend error related to `frontend.set_theme` reported here (https://github.com/home-assistant/frontend/issues/5021). I'm unsure if that error is suppressing the `themes_updated` event, which I'm assuming the companion app uses to change the status bar color.

Of note as well, it looks like the `themes_updated` event was updated to "no longer contains all theme data", so I'm wondering if that event no longer has the appropriate information to update the status bar when `frontend.set_theme` is called.

### LINKED ISSUE #645: Add option to set status bar text color in theme
In the latest release (2020.1) the option to set the status bar color by `adding app-header-background-color` to the style is implemented. However, when a light status bar color is used in the theme, the text becomes unreadable because it is always set to `UIStatusBarStyle.lightContent`. 

An option to override this, as described in https://github.com/home-assistant/iOS/issues/531#issuecomment-584289188 and https://github.com/home-assistant/iOS/issues/561#issuecomment-596089515, would help to improve the styling when using light status bars.

It might be possible to check the color of the status bar and determine whether the text color should be light or dark, but that would require a lot of work.

**Additional context**
An example of the current implementation:
IMG_2479

zacwest:
## Fix knowing about when Themes themselves change

`window.hassConnection` is `undefined` when the `.js` is initially injected, so none of the registrations were working. This adds what is effectively a runloop wait (it is available 0ms later, it's really just an ordering thing) to make sure the connection exists.

Fixes #557, fixes #645 

## Fix not having the information anymore about the Themes

This now uses JavaScript to pull the underlying css variable for the particular theme colors, which should be a less fragile method to try and get it -- rather than parsing the actual CSS itself. We only care about the currently-active values, so we don't need to parse any other ones.

## Hackily observe when the current Theme changes

This is likely to break in the future and waits on home-assistant/frontend#4220 to correctly notify us when the change occurs.

## Inspect the Theme color and set an appropriate status bar color

Tries to set the right status bar text color based on the brightness of the given background color. I like the idea of trying to solve this for all cases, rather than relying on a Theme to customize it. Open to other ideas, though.

Fixes #645.

[Video](https://i.imgur.com/9lJvO0j.mp4) of clicking a button and things happening, hooray!

### DIFF
--- a/HomeAssistant/Resources/WebSocketBridge.js
+++ b/HomeAssistant/Resources/WebSocketBridge.js
@@ -1,21 +1,39 @@
-const handleThemeUpdate = (event) => {
-  var payload = event.data || event;
-  let themeName = payload.default_theme;
-  if(themeName === 'default') {
-    window.webkit.messageHandlers.themesUpdated.postMessage({ 'name': themeName });
-  } else {
-    window.webkit.messageHandlers.themesUpdated.postMessage({ 'name': themeName, 'styles': payload.themes[themeName] });
-  }
+const notifyThemeColors = () => {
+    function doWait() {
+        let style = getComputedStyle(document.documentElement);
+        window.webkit.messageHandlers.updateThemeColors.postMessage({
+            '--app-header-background-color': style.getPropertyValue('--app-header-background-color'),
+            '--primary-background-color': style.getPropertyValue('--primary-background-color'),
+            '--text-primary-color': style.getPropertyValue('--text-primary-color'),
+            '--primary-color': style.getPropertyValue('--primary-color'),
+        });
+    }
+    // wait a short amount for the computed styles to change
+    setTimeout(doWait, 100);
 }
 
-window.hassConnection.then(({ conn }) => {
-  // window.activeHassConnection = conn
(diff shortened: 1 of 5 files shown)


HomeAssistant/Resources/WebSocketBridge.js
```diff
@@ -1,21 +1,39 @@
...
+            } else {
+                // really we just need to wait a run loop, but better safe than sorry for backoff
+                loopCount++;
+                setTimeout(doWait, loopCount * 10);
+            }
+        })();
+    });
+}
+
+waitForHassConnection().then(({ conn }) => {
```
balloob, line 30: I did not realize this code existed, but this is pretty sweet :)

TomBrien:
Looks like this fixes #645 (match text colour) too 👍

PR #652, opened 2020-06-16
Camera Notifications: Play HLS Stream

### LINKED ISSUE #649: Camera notification should stream HLS stream

When viewing the camera notification, only a low-framerate stream of JPEGs are possible. The frontend can play back HLS streams of the same camera.

Use the same mechanism to play HLS streams.

Refs #596.

zacwest:
Depends on home-assistant/core#36839. Fixes #649.

- Plays HLS streams when we're talking to a version that understands how to give us the stream address.
- Makes Camera loading multi-tier: HLS loading/failed, MJPEG loading/failed.
- Handles the media state of paused/play cleanly: pausing e.g. an HLS stream using the system UI will show the  overlay pause/play button.
- Makes Notification (Map & Camera) able to flow through promises, so we don't need to keep track of as much state.

What it looks like (fake errors from HLS to MJPEG after a few seconds): https://i.imgur.com/0AUbOcU.mp4

Fully erroring through (the last error message is meh because it was forced; initial white lag is debugger attaching): https://i.imgur.com/sJxQhwo.mp4


robbiet480:
[Bonus points](https://developer.apple.com/documentation/usernotificationsui/customizing_the_appearance_of_notifications#2359212)

PR #677, opened 2020-06-19
Refactor Sensors into smaller units

zacwest:
- Smaller (code) footprint of each sensor, in separate files.
- Tests covering all of the Sensor creations and their edge  cases.
- Easier to add new sensors into the update/registration flow; they are now all `Promise<[WebhookSensor]>`.
- Break apart the "sensor for registration" vs. "sensor for update" difference -- only the geocoded location sensor cares, and opts out where it wants.


Shared/API/Webhook/Sensors/WebhookSensor+Battery.swift
```diff
@@ -0,0 +1,70 @@
...
+            name: "Battery Level",
+            uniqueID: "battery_level",
+            icon: .batteryIcon,
+            deviceClass: .battery,
+            state: level
+        )) {
+            $0.Icon = icon
+            $0.Attributes = [
+                "Battery State": state,
+                "Low Power Mode": isLowPowerMode
```
zacwest, line 49: did this one while i was in here

PR #682, opened 2020-06-20
Fix Pedometer being deallocated during query

zacwest:
This codepath needs to maintain a strong reference to the CMPedometer, or else it deallocates which tears down the Promise prematurely, resulting in the sensors not being created/sent.

### DIFF
--- a/Shared/Common/Structs/Environment.swift
+++ b/Shared/Common/Structs/Environment.swift
@@ -189,11 +189,12 @@ public class Environment {
 
     /// Wrapper around CMPedometeer
     public struct Pedometer {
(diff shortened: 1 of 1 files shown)


Shared/Common/Structs/Environment.swift
```diff
@@ -189,11 +189,12 @@ public class Environment {
 
     /// Wrapper around CMPedometeer
     public struct Pedometer {
+        private let underlyingPedometer = CMPedometer()
```
zacwest, line 192: this is exactly the same as the CMMotionActivityManager above, not sure why i didn't do it here

PR #688, opened 2020-06-21
Change UI links to target companion.home-assistant.io

TomBrien:
This changes the UI links to target <http://companion.home-assistant.io/app/ios/**> from where they are redirected by Netlify. This means we can quickly fix any links that become dead without the need to go through review with Apple. Currently this only deals with UI links, could expand this to cover all web resources used during run time.

Counterpart PR on the companion docs repo for the redirects: home-assistant/companion.home-assistant#248

### DIFF
--- a/HomeAssistant/Views/AboutViewController.swift
+++ b/HomeAssistant/Views/AboutViewController.swift
@@ -58,7 +58,7 @@ class AboutViewController: FormViewController {
              +++ ButtonRow {
                    $0.title = L10n.About.Donate.patreon
                }.onCellSelection { _, _  in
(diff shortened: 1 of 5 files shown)


TomBrien:
Just checked these all against the preview build and all is working. Only slight caveat is I currently have "What's New" going to the doc home page.

robbiet480:
@TomBrien Can you fix the conflicts? Then this is good to merge

TomBrien:
Conflicts sorted. Going to merge the counterpart now and check the redirects then will merge this after the checks

Issue #426, opened 2019-08-10
Companion app asks for credentials unnecessarily (iPhone)


VivantSenior:
Hi,

I'm logged in to my HA instance via iPhone app for few months. I noticed that the app asks me for login and pass but after pressing the refresh button (located in the right down of the login screen) app shows my lovelace screen (but without providing credentials).
I have 4 iPhones conneted to my HA.

In my opinion app shouldn't ask for the credentials if I'm not logged out.

Best,
Viv

TomBrien:
This is a known bug. Hopefully it will be addressed in the first update to 2.0 after release. In the meantime if you force quit the app and re open, you will be logged straight in.

VivantSenior:
Thanks for the prompt response. I noticed that after "page" refresh I'm logged in again. No need for app restart.

Any timeline for v2.0?

TomBrien:
2.0 is with Apple right now I believe. We’re at the mercy of their review timeline at the moment sadly.

VivantSenior:
Thank you, finger crossed

VivantSenior:
Any updates regarding version 2.0? The Apple had time to publish 4 versions of new iOS in less than 2 weeks.. hopely they have reviewed the app.

TeNpoLe21:
Any update on this issue? I never experienced it using the betas, but just yesterday started to see it with 2019.1. Login screen is displayed and a pull to refresh brings up my front end without entering credentials. Also, force closing the app and reopening gets rid of the login prompt.

VivantSenior:
The login screen still appears on the official release but now you need to close the app and open it again to enter HA.

TomBrien:
This is a known bug. You can just pull down to refresh to fix

TomBrien:
This is now understood and will be fixed in 2019.2. Issue is related to access token expiry

TomBrien:
This is now duplicated (essentially) by #475. I'm going to close this in favour of that which is already on the 2019.2 project.

robbiet480:
@VivantSenior @TeNpoLe21 @dummptyhummpty @TravelinMax @burkemw3 Is this still a problem? I'm unable to reproduce and will need more information.

TeNpoLe21:
@robbiet480 no longer an issue for me. Thanks!

tybritten:
Same. No longer an issue

robbiet480:
Woohoo!

robbiet480:
@apateluk reports they are still having this issue.

[...]

Issue #592, opened 2020-05-25
Restoring from background sometimes gets stuck with 'Connection lost. Reconnecting…' for 10 seconds


mbaran5:
**iOS device model, version and app version**

Model Name: iPhone 11 Pro
Software Version: 13.5 
App version: 2019.1

**Home Assistant Core Version**
0.110.1 w/temporary webhook.py fix for sim2 errors 

**Describe the bug**
I receive many "Disconnected: Did not receive auth message within 10 seconds" errors from the iOS mobile app. 

```
2020-05-28 10:03:08.638 [Verbose] [main] [RegionManager.swift:243] locationManager(_:didUpdateLocations:) > didUpdateLocations [<+33.00000000,-118.00000000> +/- 65.00m (speed -1.00 mps / course -1.00) @ 5/28/20, 12:29:47 AM Pacific Daylight Time]
2020-05-28 10:03:08.639 [Warning] [main] [RegionManager.swift:268] locationManager(_:didUpdateLocations:) > Location is older than threshold.
2020-05-28 10:03:08.732 [Verbose] [main] [WebViewController.swift:584] userContentController(_:didReceive:) > getExternalAuth called
2020-05-28 10:03:08.732 [Error] [main] [TokenManager.swift:195] currentToken > Token is expired by 6405 seconds: eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJpc3MiOiIyYjZmOTkzYTg2YTc0ODhhYmViYjhlYjdiZDkwOWRhYyIsImlhdCI6MTU5MDY3NzE4MiwiZXhwIjoxNTkwNjc4OTgyfQ.KNE2ckvmVrx8myJfkAuNRc1zKfaSdsAAYR6dTCeWMh4
2020-05-28 10:03:08.885 [Verbose] [main] [WebViewController.swift:584] userContentController(_:didReceive:) > getExternalAuth called
2020-05-28 10:03:09.035 [Verbose] [main] [WebViewController.swift:584] userContentController(_:didReceive:) > getExternalAuth called
2020-05-28 10:03:09.035 [Verbose] [main] [WebViewController.swift:584] userContentController(_:didReceive:) > getExternalAuth called
2020-05-28 10:03:09.036 [Verbose] [main] [WebViewController.swift:584] userContentController(_:didReceive:) > getExternalAuth called
[...]

PR #691, opened 2020-06-21
Fix authentication delay when WebView forces a token refresh

zacwest:
Refs #592. I believe the issue is that the referenced module in the frontend is refreshing the token itself, and that's causing our expiration logic to be incorrect -- we think we have a valid token, but it's actually been replaced.

We do eventually recover when one of our own network requests yields an invalid token response, but that's _after_ we've already told the WebView about the token we believed was valid. The frontend will then retry in 10 seconds, but not before considering the token we provided as an invalid login attempt and throwing a log.

This also resolves some threading issues with the cache wrapping refresh token, whose goal was to avoid issuing more than one request at once. Since the WebView callbacks were happening on the on the main queue, and the Alamofire responses on the utility queue, it was possible for both to mutate and kick off refreshes at once, which would also log login failures.

This also adds some logs to try and understand the issues that may be happening in that ticket if this doesn't resolve them.

### DIFF
--- a/HomeAssistant/Views/WebViewController.swift
+++ b/HomeAssistant/Views/WebViewController.swift
@@ -711,9 +711,10 @@ extension WebViewController: WKScriptMessageHandler {
         } else if message.name == "updateThemeColors" {
             self.handleThemeUpdate(messageBody)
         } else if message.name == "getExternalAuth", let callbackName = messageBody["callback"] {
+            let force = messageBody["force"] as? Bool ?? false
             if let tokenManager = Current.tokenManager {
-                Current.Log.verbose("getExternalAuth called")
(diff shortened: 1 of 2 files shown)


zacwest:
This may also resolve #426.

TomBrien:
Yep I’m pretty certain they’re the same issue. I’ve seen both but since I started getting the “connection lost” error I’ve never since had the erroneous log in page.

TomBrien:
I seem to remember Paulus and Robbie dub into this on Discord ages ago. I believe the fox that was put in HA-side then got rid of the log in page issue for me and migrated it to the “connection lost” one

zacwest:
Oh that's very helpful context - yes, they're almost definitely the same issue.

PR #707, opened 2020-06-27
Minor Xcode project changes

zacwest:
- Lets Xcode run the latest migrator, disallows it to turn back on the New Build System.
- Sets the debug output as non-dSYM DWARF, so it doesn't need to create an additional file on build.
- Turns on some of the runtime sanitization in the Debug scheme.


zacwest:
Not sure why this is failing on CI.

Issue #329, opened 2019-05-05
In-app Notification Editor: category ID is forced uppercase


mgiako:
In-app Notification Editor: category identifier is forced uppercase while in the docs is specified that the category identifier MUST be lowercase.

TomBrien:
Closing this as it is fixed by #709 in that either upper or lower case is ok. Docs have been updated to reflect this

PR #709, opened 2020-06-28
Actionable Notifications: Per-Action URL, example YAML

### LINKED ISSUE #643: Support URLs in Actionable Notifications

Currently you can make a notification link to a URL, but you cannot do the same for the actionable notification buttons. Would be really useful if you could (and it does seem to be supported in the Android app). Here's an example of some forum threads asking about this, dating years back.

* [iOS Actionable Notification Call Phone Number](https://community.home-assistant.io/t/ios-actionable-notification-call-phone-number/24144) (2017)
* [Notification Actions on Local Device?](https://community.home-assistant.io/t/notification-actions-on-local-device/53613) (2018)
* [iOS Actionable Notifications, open URL?](https://community.home-assistant.io/t/ios-actionable-notifications-open-url/39481) (2018)
* [URL Notification Link](https://community.home-assistant.io/t/url-notification-link/32554/13) (2018)
* [Launching a URL using actionable notifications, is it possible?](https://community.home-assistant.io/t/launching-a-url-using-actionable-notifications-is-it-possible/78324) (2018, me haha) 
* [Make a call from actionable notification](https://community.home-assistant.io/t/make-a-call-from-actionable-notification/203476) (2020)

This feature would allow a lot more possibilities, some examples:

1. A button that calls someone (`tel://123-456-7890`)
2. A button that runs an iOS shortcut (`shortcuts://run-shortcut?name=Your Shortcut`). 
3. A button that opens a specific Lovelace view (such as `/lovelace/cameras/`).
4. Button(s) that [deep link to iOS settings](https://www.macstories.net/ios/a-comprehensive-guide-to-all-120-settings-urls-supported-by-ios-and-ipados-13-1/) (`prefs:root=SCREEN_TIME` to open Screen Time)

And a lot more...

Example use case via ashscott on forums:

example

**Additional context**

zacwest:
- Added `categoryName` to the contents of the event for activating a notification.
- Makes the acceptable case of categories both uppercase and lowercase. Refs #329.
- Copy/paste of an example service call for a category named TEST, which demonstrates per-action URLs:

```yaml
service: notify.mobile_app_#name_here
data:
  push:
    category: TEST
  action_data:
    # see example trigger in action
    # value will be in fired event

  # url can be absolute path like:
  # "http://example.com/url"
  # or relative like:
  # "/lovelace/dashboard"

  # pick one of the following styles:

  # always open when opening notification
  url: "/lovelace/dashboard"

  # open a different url per action
  # use "_" as key for no action chosen
  url:
  - "_": "http://example.com/fallback"
  - "THING": "http://example.com/url"
  - "OTHER": "http://example.com/url"
```

Fixes #643.

IMG_6666

### DIFF
--- a/HomeAssistant.xcodeproj/project.pbxproj
+++ b/HomeAssistant.xcodeproj/project.pbxproj
@@ -70,6 +70,7 @@
 		11B7FD742493225200E60ED9 /* BackgroundTask.swift in Sources */ = {isa = PBXBuildFile; fileRef = 11B7FD732493225200E60ED9 /* BackgroundTask.swift */; };
 		11B7FD752493225200E60ED9 /* BackgroundTask.swift in Sources */ = {isa = PBXBuildFile; fileRef = 11B7FD732493225200E60ED9 /* BackgroundTask.swift */; };
 		11B7FD772493232400E60ED9 /* BackgroundTask.test.swift in Sources */ = {isa = PBXBuildFile; fileRef = 11B7FD762493232400E60ED9 /* BackgroundTask.test.swift */; };
+		11C590ED24A832CA0066085D /* YamlSection.swift in Sources */ = {isa = PBXBuildFile; fileRef = 11C590EC24A832CA0066085D /* YamlSection.swift */; };
 		11C65CC0249838EB00D07FC7 /* StreamCameraResponse.swift in Sources */ = {isa = PBXBuildFile; fileRef = 11C65CBF249838EB00D07FC7 /* StreamCameraResponse.swift */; };
 		11C65CC1249838EB00D07FC7 /* StreamCameraResponse.swift in Sources */ = {isa = PBXBuildFile; fileRef = 11C65CBF249838EB00D07FC7 /* StreamCameraResponse.swift */; };
 		11CB98C6249DE15B00B05222 /* LastUpdateSensor.test.swift in Sources */ = {isa = PBXBuildFile; fileRef = 11CB98C5249DE15B00B05222 /* LastUpdateSensor.test.swift */; };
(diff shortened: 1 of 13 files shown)


vajonam:
@zacwest I just tried per-action URLs today, which isn't loading a URL, its only firing an intenral event, I am trying to use the per action url to call web hook (nabu casa) rather than hitting an internal HA url. any thoughts? should I open a bug?

zacwest:
@vajonam These URLs are for navigating the webview, not performing actions. Notification actions fire events by default; you don't need to provide URLs for executing things as a result. See the docs for [`ios.notification_action_fired`](https://companion.home-assistant.io/docs/notifications/actionable-notifications).

In your example linked above, you want to provide the `url` as a dictionary (e.g. `"url": {"action_name": "url", "action_name_two": "url2"}`) , not an array. 2020.7 out today updates the example in the app to correctly show this, if you were basing it on 2020.6's incorrect example.

vajonam:
@zacwest, thanks for the example updates, works now.

vajonam:
@zacwest  Looks like the notifications events are fired based on either the internal or external URL, what are you thoughts about adding a web-hook to to send this information to instead? that way we dont have to make my HA instance on the internet or to auto create a webhook for notification events..

vajonam:
Guess the reason I am asking this, when I am not on Wifi (HA is on not exposed externally) notifications aren't fired.

zacwest:
@vajonam The events that are fired (e.g. `ios.notification_action_fired`) are sent down via the webhook the app registers. If you are using Nabu Casa, this goes down the Cloudhook URL; if you are not, it uses the internal or external URL to reach the webhook address. Location updates behave similarly on the same webhook.

vajonam:
@zacwest, thanks for confirming this is the case. I am not seeing it use the webhook.

vajonam:
@zacwest also for completeness I am using only internal URL that `http://hass.foo.bar` which resolves internally, external URL is blank.

zacwest:
@vajonam Remote UI must be enabled for the app to use the Cloudhook URL, it mixes both of those together in a 'use cloud' preference which you cannot enable without it Remote UI.

vajonam:
@zacwest, thanks again for the reply. IMHO, this needs to be document at the very least, and I think an option for cloudhook only without remote UI is actually useful (as it was location updates) especially for actionable notifications. Probably can end the conversation here and continue it on the thread if, and if you think a FR along these lines will be entertained I will create one.

zacwest:
Yes, a feature request is the right avenue to discuss this.

PR #713, opened 2020-06-28
Fix crash on iPad trying to share example YAML

zacwest:


robbiet480:
Hate that popover shit. Gets me every time.

PR #738, opened 2020-07-04
Fix Crashlytics

robbiet480:
- Redid the Firebase config so that instead of dynamically figuring out what `GoogleService-Info-$CONFIG.plist` file to use, a build post process step copies the right file to `GoogleService-Info.plist`
- Two new debug options to trigger a crash and error that will be sent to Firebase
- Improved dSYM uploading
- Enabled `-FIRDebugEnabled` argument in Debug
- Re-enabled XCGLogger to send to Crashlytics. Logs will only be uploaded for a error or crash
- Set HA_Version on Crashlytics again


HomeAssistant/AppDelegate.swift
```diff
@@ -892,24 +892,9 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
     }
 
     func setupFirebase() {
-        var fileName = "GoogleService-Info-Debug"
+        LogDestination = CrashlyticsLogDestination()
```
zacwest, line 886: could do…

```suggestion
        Current.Log.add(CrashlyticsLogDestination())
```

robbiet480: I don’t think that’ll work since `Log` is an initialized variable...

PR #751, opened 2020-07-05
CI: Skip pod install if already up-to-date

zacwest:


zacwest:
Not worth it without the Manifest.lock check, the spec repo isn't saved locally (of course) and caching many GBs of spec repo feels like a bad idea since it does a shallow fetch

PR #775, opened 2020-07-11
Fix off-main-queue crash during invalid connection in beta

zacwest:
UIAlertController doesn't want to be called off the main thread, and I did not intend for this retry behavior to jump to the global queue.

### DIFF
--- a/Shared/Networking/ConnectionInfo.swift
+++ b/Shared/Networking/ConnectionInfo.swift
@@ -210,32 +210,34 @@ public class ConnectionInfo: Codable {
         Current.Log.error(errMsg)
 
         #if os(iOS)
-        let alert = UIAlertController(title: "URL Unavailable",
-                                      // swiftlint:disable:next line_length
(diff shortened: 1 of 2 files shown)


Shared/Networking/ConnectionInfo.swift
```diff
@@ -210,32 +210,34 @@ public class ConnectionInfo: Codable {
...
-            pTextField.placeholder = self.activeURLType.description
-            pTextField.clearButtonMode = .whileEditing
-            pTextField.borderStyle = .none
-            textField = pTextField
-        }
+        DispatchQueue.main.async {
+            let alert = UIAlertController(title: "URL Unavailable",
+                                          // swiftlint:disable:next line_length
+                                          message: "Expected to have a \(self.activeURLType) but none available! Please enter the URL. App will exit after entry, please reopen.",
+                                          preferredStyle: .alert)
```
robbiet480, line 217: Should we put these strings into Lokalise?

zacwest: yeah, probably - but let's maybe take a look at how we deal with this alongside websocket work so we can maybe avoid needing to care, too

PR #785, opened 2020-07-15
Add additional monitored regions to zones <100m 

### LINKED ISSUE #784: Zone enter events with radius <100m send locations outside zone
App version: 2020.3 w/setting, 2020.4+

**Describe the bug**
iOS sends enter events when 100m from the center of a monitored region, which means we send a location event too soon. 2020.3 (without the in-dev setting) and prior send the zone center location for all enter events, which masks this issue.

**To Reproduce**
Create a zone with a radius <100m and note the location updates outside the zoe.

**Expected behavior**
Location updates are sent when actually in the zone.

**Additional context**

A few possible solutions:

1. Draw zones around smaller zones that overlap and are centered about 100m from the edge of the zone, so we get notified when entering/exiting these secondary zones. We don't really care about their state, we just want to be _notified_ of them.
1. Mutate the location sent for exit events to always be within the zone. Instead of sending the center, draw a straight line from the outer circle of the zone and move the location as close as necessary.

zacwest:
Monitored regions that are too small fire enter events when the user is far outside. This mitigates the problem by adding additional monitored regions around smaller ones that can fire enter and exit events when the user is actually inside the smaller one.

Fixes #784.

For example, if I've got a zone centered around South Park at 50m…

unknown

- Instead of the requested zone (green) it will create regions that overlap it and are 100m in radius each.
- For the purposes of determining inside or outside zone, the original (green) zone's information is used.
- For the purposes of deciding when to decide to send a location update, the additional (red) zone's statuses are used.


zacwest:
Punting on trying to filter down the monitored region count for now, instead adding a log and an error report so we can see what it's like in the wild. Problems here:

1. Beacons are also limited to 20, but we don't have a good sense of _which_
2. We may want to do location-based filtering of regions, but we don't necessarily know where the user is

These will require more work on our part to figure out, and that can happen later

PR #827, opened 2020-07-24
Add Perform Action intent, donate intents & suggestions

zacwest:
Adds a new Shortcut action called "Perform Action" which performs the given action. The parameter to this Shortcut requires the user choose from a list which action to execute.

This is donated as an intent every time an action is performed, and this intent is flagged as suggest-able, so it'll start appearing on lock screen, search, and in the Shortcuts app.

This also provides every possible action as a suggestion, which causes them to appear under the "Perform Action" bucket even if they haven't been executed (recently or otherwise).

Image 10


TomBrien:
This looks awesome. Donatable intents is a nice addition.

PR #834, opened 2020-07-25
"Add to Shortcuts" button in Action configurator

zacwest:
- Restores the cancel button for Action editing.
- Adds the Add to Shortcuts button in the "Name" section.

Image 12


TomBrien:
Does this register `name` or `text` as the Siri command?

zacwest:
I left the "suggested phrase" option blank, so it defaults to an empty field to type in. I'd imagine if we wanted to suggest a default that "text" might be the right one.

TomBrien:
I think leaving empty sounds perfect actually.

PR #838, opened 2020-07-27
Allow specifying Scene Action colors via Scene attributes

zacwest:
Adds parsing for `icon_color`, `text_color` and `background_color` as read from the Scene. Whenever these are set in the config, we hide the corresponding row when configuring its action.

Image 14


TomBrien:
Might be worth having a string "The appearance of this action is defined by the scene configuration in Home Assistant" similar to the "You can also change these by customizing…"

PR #842, opened 2020-07-28
Generate more compatible random colors for Actions

### LINKED ISSUE #422: Simplify New Action screen
(please label 2.0 Beta)

There are two parts of this screen that I think could be improved on after adding many actions:

1. `Action` as the default value in the "Text" field. This has to be deleted and replaced with a value. Perhaps it could act more like a placeholder so it doesn't need to be deleted?

2. The colour controls of the text/background/icon are handy but the randomisation means I have to do a lot of scrolling/tapping to get mine all looking the same. Perhaps there should be black text/white background defaults, leaving the user to set custom colours if they want to?

IMG_2576

zacwest:
Fixes #422. This aims to create background colors that are:

1. Any hue (e.g. red, green, blue)
2. Saturation between 50% and 100% (avoiding super gray colors)
3. Brightness between 25% and 75% (avoiding blacks and whites)

It'll then set both the text and icon color to the same pairing we use for calculating whether to use a light or dark status bar color, sticking to white or black.

Examples:

Image 16
Image 17
Image 18

### DIFF
--- a/HomeAssistant/Views/ActionConfigurator.swift
+++ b/HomeAssistant/Views/ActionConfigurator.swift
@@ -147,13 +147,10 @@ class ActionConfigurator: FormViewController, TypedRowControllerType {
                 $0.value = self.action.IconName
             }.cellUpdate({ (cell, row) in
                 if let value = row.value {
-                    let theIcon = MaterialDesignIcons(named: value)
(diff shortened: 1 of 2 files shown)


zacwest:
My logic seems to avoid light enough colors that cause the text to go black. I think I'm okay with that.

PR #846, opened 2020-07-29
Minor FCM improvements

robbiet480:
Based on learnings during the account migration, this PR does two things:

1. Subscribes users to the following FCM topics:

- `everyone`
- `app_version_<VERSION_NUMBER>` e.g. `app_version_2020.4.1`
- `app_version_build_<VERSION_NUMBER>_<BUILD_NUMBER>` e.g. `app_version_build_2020.4.1_1`
- `ha_version_<HA_VERSION>` e.g. `ha_version_0.112.1`
- `locale_<USER_APP_FIRST_LOCALE>` e.g. `locale_en-US`

This way, we can target users via topic instead of having to use the limited functionality FCM composer.

2. If `fcm_options.image` is set in a notification payload, use that as the attachment, ignoring anything else. This way, if we do have to use the limited functionality FCM composer we can still set an image.

### DIFF
--- a/HomeAssistant/AppDelegate.swift
+++ b/HomeAssistant/AppDelegate.swift
@@ -826,6 +826,7 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
         Lokalise.shared.localizationType = Current.appConfiguration.lokaliseEnv
     }
 
+    // swiftlint:disable:next function_body_length
     func setupFirebase() {
         LogDestination = CrashlyticsLogDestination()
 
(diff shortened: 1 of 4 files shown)


HomeAssistant/AppDelegate.swift
```diff
@@ -858,6 +859,46 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
...
+        }
+
+        let newTopics = [
+            "everyone",
+            "app_version_\(Constants.version)",
+            "app_version_build_\(Constants.version)_\(Constants.build)",
+            "locale_\(Bundle.main.preferredLocalizations.first ?? "UNKNOWN")"
+        ]
+
+        for topic in newTopics {
```
zacwest, line 892: it wouldn't be too bad to use Set logic to do diffing, e.g. in

https://github.com/home-assistant/iOS/blob/master/Shared/API/Models/ModelManager.swift#L149-L153

robbiet480:
Closing as we are moving off FCM

PR #857, opened 2020-08-01
Improve Connection Settings saving and editing

### LINKED ISSUE #717: URL Internal / External not storing user input

Model Name: iPhone 11
Software Version: 13.5
App version: 2020.3 (4)

0.111.1

**Describe the bug**

URL Internal / URL External, values are not stored sometimes, may be related to the app for some reason not being able to reach the server or for unknown reason, the app seems to fall back to an older value thus deleting what you enter manually, this just confuses and gets you frustrated

My external url is properly hair pined, I'm behind two nats, but all port forwarding are setup properly.

Also, I should add that onboarding is not great for me, it is not detecting my home assistant installation, I'm not sure why. I'm using many ESPHome nodes which use mDNS discovery and I have no problem with those.

**To Reproduce**

**Expected behavior**
Store the manually entered value, even if it's wrong, it will be awesome to get a popup with an error (unresolved host, unexpected response, timeout, whatever)

**Screenshots**

**Additional context**
I reported this earlier, but got closed. https://github.com/home-assistant/iOS/issues/425

Here is another report 

https://github.com/home-assistant/iOS/issues/539

May be related: https://github.com/home-assistant/iOS/issues/459

zacwest:
Moves setting an internal or external URL into a sub-screen which has cancel/save buttons and an explicit lifecycle around whether the save is successful:

- When successful, pops off and shows the new value.
- When unsuccessful for connectivity reasons, allows the user to ignore the error and persist the URL change anyway.
- When unsuccessful because we reject the URL as being Nabu Casa or trying to clear the last one, we tell the user we can't save.

This also allows clearing external if internal is set and fixes trying to reset to external when there is no external.

Improves how and when we persist SSIDs in the list to be the same save button next to internal URL and adds a warning if location permission suggests we won't be able to determine SSID information.

Fixes #717.

Image 22
Image 23

### DIFF
--- a/HomeAssistant.xcodeproj/project.pbxproj
+++ b/HomeAssistant.xcodeproj/project.pbxproj
@@ -33,6 +33,7 @@
 		1141182A24AFA10900E6525C /* WebhookResponseHandler.swift in Sources */ = {isa = PBXBuildFile; fileRef = 1141182924AFA10900E6525C /* WebhookResponseHandler.swift */; };
 		1141182B24AFA10900E6525C /* WebhookResponseHandler.swift in Sources */ = {isa = PBXBuildFile; fileRef = 1141182924AFA10900E6525C /* WebhookResponseHandler.swift */; };
 		114FACAE24B2ABA2006C581F /* Promise+WebhookJson.test.swift in Sources */ = {isa = PBXBuildFile; fileRef = 114FACAD24B2ABA2006C581F /* Promise+WebhookJson.test.swift */; };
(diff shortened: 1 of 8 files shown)


pjaspers:
Thanks, I think I'll need this :)

Moved homeassistant to a different IP, but now it keeps falling back to the old IP as soon as one of the network requests fails, thus rendering the app unusuable because if I change it back to the correct IP I get a 'Error: Response could not be serialized, input data was nil or zero length'

PR #863, opened 2020-08-03
NFC Tag Reading, Writing & Automatic Event Firing

zacwest:
This is the iOS pair to home-assistant/android#689. This introduced a bunch of things:

1. Adds handling for Universal Link-based NFC tags with URLs like `home-assistant.io/nfc/<anything>` where  `<anything>` is used to fire an event like:

```
["event_type": "tag.read", "event_data": ["tag": "<anything>", "sourceDevicePermanentID": "…", "sourceDeviceName": "…", "sourceDeviceID": "…", ]]
```

2. Adds an NFC Tags integrations/settings screen, which has a few features:

- Read Tag. This intentionally can only read the tags with values like the above. When read, it allows duplicating (e.g. writing the same identifier to a new tag), firing the event associated (for testing), and shows an example trigger for the event.
- Write Tag. This either creates a random identifier, or allows the user to enter a manual identifier. This opens afterwards the same screen as the read flow.

Image 24
Image 26


David-Development:
@zacwest I just noticed that you used `sourceDeviceName` in the event data. Shouldn't we stick to the snake case convention? Otherwise the snake and camel case convention is mixed. So it would be `source_device_name` or even shorter.. just `device_name`?

zacwest:
> @zacwest I just noticed that you used `sourceDeviceName` in the event data. Shouldn't we stick to the snake case convention? Otherwise the snake and camel case convention is mixed. So it would be `source_device_name` or even shorter.. just `device_name`?

iOS has historically used these camel case keys for the device information for events it sends for actions and actionable notifications. I think one way we could move towards unification (snake case is okay with me) would be to send both versions, but this might require a bit of retooling to how we show the example triggers.

David-Development:
It probably would make sense to unify these events a little more but I see the problems that arise from it. It's definitely not a great solution if the user had to place both cases in the trigger. Since probably not many users have adopted the new nfc on ios feature yet maybe it isn't too late to change it to snake case? Otherwise I would also be okay with the camel case.

PR #890, opened 2020-08-11
iOS 14 Widget: Actions

zacwest:
Adds a new widget: "Actions" which can hold some number of actions: small (1), medium (1-4), large (1-8).Each action displays based on the text, icon, and background configuration that already exists.

When tapped, the actions launch the app and perform the action. This is an Apple restriction - Widgets cannot execute code when interacted with; they can only launch the app when tapped. Apple gives special permission to their Shortcuts widgets to change this behavior, of course, because all developers are equal but some are more equal than others.

I took "inspiration" from the general layout of most Widgets, but I really don't like the wasted space that Apple does with some of their widgets. For example, the Shortcuts widget draws margin after margin after margin around the actions. This does not. They also force you to have a bunch of blank ones if you don't have exactly the number they want. This fills as best as possible, even if it means making widgets (gasp!) larger than the smallest possible size.

Image 2
Image 3


Nickno95:
I understand not liking the wasted space, but I think the margins around the buttons in the shortcuts widgets make them seem cleaner and more like buttons. Is it possible to have the margins as an optional toggle? At least to see how it looks.

PR #954, opened 2020-08-30
Add BatterySensor Update Signaler

zacwest:
Adds an Update Signaler for battery state (but not level) changes. Level is effectively a timer for updating (on Mac it's every 60 seconds), which is effectively the same as our periodic timer -- it doesn't signal any kind of state changes, just that time has elapsed. This will, for example, update sensors immediately upon charge starting or ending (while the app is running, of course). This works on macOS and iOS but not on watchOS.


c0nt4x:
Will the app need to be opened or is it sufficient to be running in the background?

zacwest:
The iOS app will not live update if it isn't in the foreground. The Mac app will.

PR #968, opened 2020-08-31
"Active" Sensor for Mac

zacwest:
This binary sensor reports `on` when the machine is active, `off` otherwise. Active state tracks: 

- Idle time, defaulting to 5 minutes. If no keyboard, mouse, or tablet (wat) input is detected in that period, we consider this idle. This is exposed as a Sensor setting (in Preferences > Sensors > Active).
- Screensaver starting and ending.
- Sleeping starting and ending.
- Screen turning off and on (also called sleeping in some places in Apple-land).
- Lock Screen, which is both the Apple > Lock Screen and the Fast User Switching mechanism.
- Fast user switching. This is the Menu Bar "switch to X user" and can also end up triggering Lock Screen too.

If none of these are happening, we consider the machine active.

image


thenickdude:
Is it possible to allow a "time until idle" of "0" to mean "don't track idle state"? 

It's really annoying when I'm watching a video and my desk lights turn off because Home Assistant thinks my Mac is "idle", even though YouTube is keeping the screen on for me due to a video playing.

PR #981, opened 2020-09-03
Enable server-defined actions for 0.115

### LINKED ISSUE #494: Possiblity to edit actions trough a file/interface from PC
Hi,

As some who uses Actions a lot, i find the process off fiddling on my mobile device to add them, color them etc. a bit hard. Would it be possible to for example use a .yaml file and load them in just like you load notification categories?

This way its way easier to for example share them with multiple Iphone + Watch users.

TomBrien:
Per home-assistant/core#38572 we can define actions in YAML. Actions are provided as view due to HA limitations in getting one integration (`ios` here) to access webhooks registered to another (`mobile_app`).

Tested locally in the following cases:
- `ios:` loaded in `configuration.yaml` with `actions` key.
- `ios:` loaded in `configuration.yaml` without `actions` key.
- `ios:` not loaded.

Push categories are also loaded fine.

Fixes #494

### DIFF
--- a/Shared/API/HAAPI.swift
+++ b/Shared/API/HAAPI.swift
@@ -401,38 +401,23 @@ public class HomeAssistantAPI {
     }
 
     public func GetMobileAppConfig() -> Promise<MobileAppConfig> {
-        if Current.serverVersion() < .actionSyncing {
-            return firstly { () -> Promise<MobileAppConfigPush> in
-                requestImmutable(path: "ios/push", callingFunctionName: "\(#function)")
-            }.recover { error -> Promise<MobileAppConfigPush> in
-                if case AFError.responseValidationFailed(reason: .unacceptableStatusCode(code: 404)) = error {
-                    Current.Log.info("ios component is not loaded; pretending there's no push config")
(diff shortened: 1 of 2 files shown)


zacwest (approved):
Thanks for bringing this across the finish line!

Since the code paths for both are now going through a REST call, let's merge them together like…

```swift
        return firstly { () -> Promise<MobileAppConfig> in
            if Current.serverVersion() < .actionSyncing {
                let old: Promise<MobileAppConfigPush> = requestImmutable(
                    path: "ios/push",
                    callingFunctionName: "\(#function)"
                )
                return old.map { MobileAppConfig(push: $0) }
            } else {
                return requestImmutable(path: "ios/config", callingFunctionName: "\(#function)")
            }
        }.recover { error -> Promise<MobileAppConfig> in
            if case AFError.responseValidationFailed(reason: .unacceptableStatusCode(code: 404)) = error {
                Current.Log.info("ios component is not loaded; pretending there's no push config")
                return .value(.init())
            }

            throw error
        }
```

zacwest (approved):

PR #1026, opened 2020-09-16
Fix path to resource file for manual licenses

zacwest:


zacwest:
Got all the way to the end of the archive before it realized.

PR #1097, opened 2020-09-25
Assign a different bundle identifier for Shared.framework on watchOS

zacwest:
When the bundle identifier of an iOS and watchOS framework match, the Swift Package Manager integration in Xcode has difficulties separating them out, so it ends up erroring in strange ways. In our case, this looks like `error: no such module 'Clibsodium'` failures at build time for one of them.

We don't need the bundle identifier to match -- in fact, it doesn't matter either way, since we don't need to code sign it, and we never reference it directly


zacwest:
This fixes a sample project but doesn't fix HA. 😢

PR #1153, opened 2020-10-04
Really fix the example notification category YAML

zacwest:
Finishes #1148.


Sources/Shared/API/Models/NotificationCategory.swift
```diff
@@ -99,7 +99,7 @@ final public class NotificationCategory: Object, UpdatableModel {
           # use "\(Self.FallbackActionIdentifier)" as key for no action chosen
           url:
             "\(Self.FallbackActionIdentifier)": "http://example.com/fallback"
-            \(urlStrings.joined(separator: indentation + "- "))
+            \(urlStrings.joined(separator: indentation + "  "))
```
TomBrien: Any reason to not add the two spaces to `indentation` on L77. Doesn't seem to be used anywhere else

PR #1197, opened 2020-10-16
Support multiple Apple Watch complications

### LINKED ISSUE #839: Allow creating multiple complications per type
We can now create multiple complications per type, so we'll need to evolve the Watch configurations at some point to support this.

zacwest:
This allows for configuring more than 1 e.g. Modular Small at a time. This requires watchOS 7; for releases before watchOS 7 we do not allow a secondary template to be created. This could be improved by allowing swapping between one of the multiple for watchOS 6 and earlier, but I didn't implement this.

- Adds a "Display Name" to Complications. This is optional and falls back to the template type. This is primarily to differentiate complications; the underlying unique ID is just a UUID.
- Enable previews of complications, which is used when selecting them. This is a stripped-down version of the items in the template, usually just an icon provider or one of the text providers. This is largely easier now since we're now requesting templates for all complications, even those not actively visible.

This also changes the flow when setting up initially. Instead of a list of families to choose from at the start, the user must tap the 'Add' button and pick one from there.

Fixes #839.

Image
Image


mountainsandcode:
Amazing work on the watchOS features recently - thanks a ton, @zacwest!

PR #1223, opened 2020-10-21
Move Sentry into Shared, add to all extensions

zacwest:
This gives us crash reporting for extensions that run on the phone and watch.


zacwest:
I think the setup is running too early and uploading crashes isn't working as a result.

zacwest:
I take that back, it is reporting - Sentry's UI doesn't make it super obvious how to find them. Example crashes with this move:

- [Intents](https://sentry.io/organizations/home-assistant-io/issues/1971287546)
- [Main App](https://sentry.io/organizations/home-assistant-io/issues/1971220995)

PR #1253, opened 2020-10-29
Allow hiding complications when watch is locked

TomBrien:
Given that we essentially let users add anything from Home Assistant that can be represented as text as a complication I thought it would be a good idea to let people choose whether or not to display a complication when the watch is locked.

Tested on my watch and seems to be working 😯although I haven't tested the Realm migration yet. Apologies in advance, this is my first time trying something of this complexity (for me, appreciate it's actually pretty simple)

### DIFF
--- a/Sources/App/Settings/AppleWatch/ComplicationEditViewController.swift
+++ b/Sources/App/Settings/AppleWatch/ComplicationEditViewController.swift
@@ -147,6 +147,11 @@ class ComplicationEditViewController: FormViewController, TypedRowControllerType
             $0.value = self.config.name
         }
 
+        <<< SwitchRow {
+            $0.title = L10n.Watch.Configurator.Rows.IsPrivate.title
+            $0.value = self.config.IsPrivate
+        }
+
         <<< PushRow<ComplicationTemplate> {
             $0.tag = "template"
             $0.title = L10n.Watch.Configurator.Rows.Template.title
--- a/Sources/App/Settings/AppleWatch/ComplicationEditViewController.swift
+++ b/Sources/App/Settings/AppleWatch/ComplicationEditViewController.swift
@@ -147,11 +147,6 @@ class ComplicationEditViewController: FormViewController, TypedRowControllerType
             $0.value = self.config.name
         }
 
-        <<< SwitchRow {
-            $0.title = L10n.Watch.Configurator.Rows.IsPrivate.title
-            $0.value = self.config.IsPrivate
-        }
-
         <<< PushRow<ComplicationTemplate> {
             $0.tag = "template"
             $0.title = L10n.Watch.Configurator.Rows.Template.title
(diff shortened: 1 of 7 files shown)


TomBrien:
Hmm I'm getting:
> Database Error
> Migration is required due to the following errors:
> - Property 'WatchComplication.IsPrivate' hass been added.

Pop up. Not sure what I did differently to https://github.com/home-assistant/iOS/blob/master/Sources/Shared/Common/Extensions/Realm%2BInitialization.swift#L89-93

TomBrien:
Migration should be resolved now

Sources/App/Settings/AppleWatch/ComplicationEditViewController.swift
```diff
@@ -184,6 +189,11 @@ class ComplicationEditViewController: FormViewController, TypedRowControllerType
             cell.detailTextLabel?.text = row.value?.style
         }
 
+        <<< SwitchRow("IsPrivate") {
+            $0.title = L10n.Watch.Configurator.Rows.IsPrivate.title
+            $0.value = self.config.IsPrivate
```
zacwest: I think it is worth flipping the way this is presented. "Show When Locked {default on}" is a little easier to mentally discern than the double-negative "Hide when locked {default off}".

Sources/Extensions/Watch/ComplicationController.swift
```diff
@@ -41,7 +41,25 @@ class ComplicationController: NSObject, CLKComplicationDataSource {
         for complication: CLKComplication,
         withHandler handler: @escaping (CLKComplicationPrivacyBehavior) -> Void
     ) {
-        handler(.showOnLockScreen)
+
+        let model: WatchComplication?
+
+        if #available(watchOS 7, *), complication.identifier != CLKDefaultComplicationIdentifier {
```
zacwest: given this is from another method in this class, i think it's worth writing a helper function like `func complication(for complication: CLKComplication) -> WatchComplication?`

zacwest (changes requested):
overall nice job! just a couple suggestions.

TomBrien:
Had a bit of issues getting complications to sync. Should note I've seen this since the restructuring of complications for watchOS 7. I think it may be a me thing but interested if can be reproduced by others. Complication privacy is still working and tested though. Bottom complications are both HA left is set with Show on When Locked true right has it false. 
Unlockedimage
Locked image

Sources/Extensions/Watch/ComplicationController.swift
```diff
@@ -16,8 +16,8 @@ class ComplicationController: NSObject, CLKComplicationDataSource {
     // https://github.com/LoopKit/Loop/issues/816
     // https://crunchybagel.com/detecting-which-complication-was-tapped/
 
-    private func template(for complication: CLKComplication) -> CLKComplicationTemplate? {
-        Iconic.registerMaterialDesignIcons()
+    private func getComplicationModel(for complication: CLKComplication) -> WatchComplication? {
```
zacwest: prefixing functions with 'get' is not really a swift-ism that works super well -- apple's been leaning on doing this only for asynchronous returns (e.g. it provides a block)

Sources/Extensions/Watch/ComplicationController.swift
```diff
@@ -41,7 +51,16 @@ class ComplicationController: NSObject, CLKComplicationDataSource {
         for complication: CLKComplication,
         withHandler handler: @escaping (CLKComplicationPrivacyBehavior) -> Void
     ) {
-        handler(.showOnLockScreen)
+
+        let model: WatchComplication?
+
+        model = getComplicationModel(for: complication)
```
zacwest: if you pull this up into the variable declaration above you can avoid needing to declare the type, too

zacwest (approved):
just minor - let's try and keep to swift-isms for function names here

PR #1284, opened 2020-11-19
Add "Placeholder" complication

### LINKED ISSUE #525: Add app icon as a complication image or template
If you open a complication whcih defaults to the app icon (such as Graphic - Circular) in the "Apple Watch" setting in the app, the icon is replaced by the default icon set for that (i.e. WiFi connection) and it is not possible to revert to the (nicer) app icon. Can a template be added to the relevant complications (all those that use icons) that just uses the app icon? At the moment the only way to recover the icon is to reset the companion app

### LINKED ISSUE #1281: iOS HA Complication not showing in list
Model Name: 
Software Version: iOS 14.1 / Watch OS 7.1
HA App version: 0.117.6

Home Assistant 0.117.6

**Describe the bug**
The HA complication isn't showing on the apple watch. The complication I'm referring to is the one that shows all the actions you have setup. I've tried uninstalling/resetting/reinstalling the app on the connected iPhone.

zacwest:
## Summary
Since we only offer complications for configured versions, we don't have a quick "launch the app" complication available without creating one, and you cannot create one that's just the HA logo.

Rather than adding a new template configuration for the app icon, this adds a complication to all places we support which show some variant of "Home Assistant", "HA" and the app icon without any kind of "this is an error!" state.

Fixes #1281. Fixes #525.

## Screenshots
image
image

### DIFF
--- a/Sources/Extensions/Watch/ComplicationController.swift
+++ b/Sources/Extensions/Watch/ComplicationController.swift
@@ -78,8 +78,15 @@ class ComplicationController: NSObject, CLKComplicationDataSource {
         if let fromComplication = self.template(for: complication) {
             template = fromComplication
         } else {
-            Current.Log.error("failed to get template, providing error")
-            template = ComplicationGroupMember(family: complication.family).errorTemplate
+            Current.Log.info("no configured template, providing placeholder")
+
+            if #available(watchOS 7, *) {
+                template = ComplicationGroupMember(family: complication.family)
+                    .fallbackTemplate(for: complication.identifier)
+            } else {
+                template = ComplicationGroupMember(family: complication.family)
+                    .fallbackTemplate(for: nil)
+            }
         }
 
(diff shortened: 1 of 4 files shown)


TomBrien:
🎉

Sources/Extensions/Watch/ComplicationController.swift
```diff
@@ -78,8 +78,15 @@ class ComplicationController: NSObject, CLKComplicationDataSource {
         if let fromComplication = self.template(for: complication) {
             template = fromComplication
         } else {
-            Current.Log.error("failed to get template, providing error")
-            template = ComplicationGroupMember(family: complication.family).errorTemplate
+            Current.Log.info("no configured template, providing placeholder")
```
TomBrien, line 81: Do these remain available when a complication is configured for that family?

zacwest: on watchOS 7, yes. on watchOS 6, no.

PR #1288, opened 2020-11-21
Local Push Provider

zacwest:
Mega WIP.

Major themes:

- [ ] Finish up WebSocket connectivity, use it here.
- [ ] Handle WebSocket -> push.
- [ ] Handle entitlements, since I had to hard-code developer ones here; automatic doesn't work with this unsurprisingly.


inthami:
Hi,
Is it possible that after invoking actions from widgets on macOS, the action will activate without moving the user to the Home assistant app? The confirmation could appear briefly in the widgets. This is a bit annoying.
Thank you very much

zacwest:
Cleaning up old PRs. Still doing, just don't like this floating around…emotionally.

PR #1307, opened 2020-12-20
Fix crash when Map notification is provided invalid latitude/longitude

zacwest:
- Fixes a crash when `latitude` or `longitude` are String types but not able to be converted to Double/CLLocationDegrees.
- Adds support for Int or Float-based `latitude` or `longitude`, e.g. what native templates would provide were it not currently broken.


TomBrien (approved):
👍

PR #1330, opened 2021-01-02
Big Sur versions of the Mac icon

zacwest:
Fixes the icon being full-bleed without any rounding, now that there's a Mac-specific one. Bumps build.

Using the Sketch template [from Apple](https://developer.apple.com/design/resources/#macos-apps).


zacwest:
The builds will continue until morale improves.

PR #1355, opened 2021-01-07
macOS Big Sur custom icons

XusBadia:
## Summary

Since Apple introduced a new design in the latest version of macOS Big Sur, I designed two new icons that follow the new icon design.

## Screenshots

image
image


zacwest:
The white border feels a bit off to me, compared to the general Big Sur icons. What do you think @balloob? I think @TomBrien mentioned that the shadow may outside desired brand things.

XusBadia:
> The white border feels a bit off to me, compared to the general Big Sur icons. What do you think @balloob? I think @TomBrien mentioned that the shadow may outside desired brand things.

Thanks! Here's a version without the border.

image

What you mention about the shadow is interesting... I've never read anything related to that, but you may be right. Let's see what they say.

balloob:
I don't really see anything wrong with our current logo, and it fits right in?

Screenshot 2021-01-13 at 10 18 39@2x


The shadows make the nodes look "dirty". I don't like it.

XusBadia:
> I don't really see anything wrong with our current logo, and it fits right in?
> 
> Screenshot 2021-01-13 at 10 18 39@2x
> 
> The shadows make the nodes look "dirty". I don't like it.

I do find it a bit too different when compared to the other icons, but that might be just personal preference.

zacwest:
@XusBadia thank you for the contribution. I think we can iterate through and get something that matches the new Big Sur icon a little closer, if you'd like to play around. I think in general, the difference right now for us on Big Sur is:

- The home iconography doesn't quite match the icon grid of Big Sur icons
- There's an added 3d to many macOS icons which the subtle shadow of the current icon doesn't address
- There's a _lot_ of blue icons on macOS, and this one feels like it may get lost in the pile a bit

I tried out your current version in the app (you can just paste it when getting info on a .app so it's easy) and I think it's not bad, but the shadow affect on the branches doesn't help as much. What do you think?

PR #1376, opened 2021-01-16
Add server-driven alerts based on client or server versions

zacwest:
## Summary
Shows an alert based on an alerts site JSON array, allowing us to customize alerts based on client or server version.

## Screenshots
Screen Shot 2021-01-16 at 14 06 40Screen Shot 2021-01-16 at 14 06 24
Image 4

Not quite documentation, but the server-side JSON for this is here: home-assistant/alerts.home-assistant.io#269

## Any other notes
- Does not send any local information anywhere. This only pulls the JSON remotely.
- Dismissing (by tapping the background, or swiping down) or opening the alert prevents it from displaying again.
- Supports any number of alerts, showing only the relevant and unseen ones.


mountainsandcode:
Hi Zac! 

This is an excellent addition for the average user and will also be very helpful to assist people to keep their iOS app version up-to-date when big things are happening!

One question/request: What happens on a device that doesn’t have a live internet connection like a iPad as a kiosk? No error message I hope? Can we add a setting to disable this automatic web request for power users who know about this anyway?

zacwest:
Failing to load doesn't show anything. I'll look into adding a general-purpose privacy setting, but it's worth noting (again, in case other people just read the comment here) that nothing is sent up with the request, it pulls down only.

PR #1385, opened 2021-01-18
Allow showing in the menu bar, either additionally or exclusively

### LINKED ISSUE #949: MacOS app – ability to move app to the Menu Bar
Updating sensor does not work when app is turned off. On the other hand app does space in the Dock (additional icon) when you don't use it. You need to remember to launch it manually as well, and also remember to not close the window but minimize it instead – otherwise app would be terminated. [That last one is already solved.]

Actions are not quickly accessible too – I need to open app window and only then I can click 'Actions' on Menu Bar and select one. That's at least 3 clicks and a lot of eye and mouse movement.
 
1. I'd like to be able to move this app to Menu Bar and keep it there (maybe some setting – Dock / Menu Bar / both). Actions could be easily triggered from there also – just two clicks, even if they would be nested in some submenu.
2. An option to make app autostart (on system boot / user login) would be highly appreciated.
3. Closing app window shouldn't close the app. [That's already solved]

If moving an app from the Dock entirely is not possible with Catalyst app, it maybe can exist both in the Dock and as a Menu Bar icon – for easy access to actions.
Also, maybe there's a possibility to update sensors even if app is not running.

zacwest:
Fixes #949 (the menu bar part).

## Summary
Allows configuring 2 potentially-mutually-exclusive things: whether to show the app in the dock, and whether to show a status item in the menu bar. Wraps them into one setting.

## Screenshots
Screen Shot 2021-01-17 at 20 55 54Screen Shot 2021-01-17 at 20 56 03
Screen Shot 2021-01-17 at 20 56 19Screen Shot 2021-01-17 at 20 56 26

Worth noting there seem to be something like 16 different permutations for the look of the menu bar. Here's some more!

imageimage

## Any other notes
- Positions the status item handling (routed through the Mac Bridge) in the same place that we handle menus on Catalyst, since the ultimate fate of the status item will be offering menus of some sort.
- Clicking on the status item hides or shows the app, depending on the current state. Effectively acts like a toggle. If there are no windows open when the app needs to be activated, creates a new webview window.
- Doesn't allow disabling both the menu bar item and the dock item, since the app isn't really set up for success with that.


oscar-b:
@zacwest How do I get to the app preferences after enabling Menu Bar mode?

jasoncodes:
I got stuck for a bit trying to get into Preferences again too. I expected a context menu on the menu bar icon.

Try `⌘,`.

oscar-b:
> I got stuck for a bit trying to get into Preferences again too. I expected a context menu on the menu bar icon.
> 
> Try `⌘,`.

Ah cool, thanks, that allows me to get back to normal again :)

zacwest:
Oh interesting. This is good feedback and one of those things that as the author I didn't even notice would be confusing.

When you tried to get a context menu, did you right-click the menu bar item at all? I've got a menu system working locally right now but I was debating whether to include it in left or right click. Left is way more standard but makes it slower to re-open the app.

oscar-b:
I tried right-clicking the menu bar. I also went looking for the preferences in the normal places, where I think the iOS version displays it?

Just the fact that menu bar only opens a window which doesn't have a menu in macOS was a little confusing. I don't think that's very common? Only for stuff that disappear when not focused anymore, like a calendar widget or something?

jasoncodes:
The first thing I tried after enabling the menu bar item was to right click it. I like being able to quickly toggle with left click. I feel like this behaviour is common enough to not be confusing. A few apps I use regularly that perform an action or toggle visibility on left click and show a menu on right click are Due, Bartender, ToothFairy, and Fluid.

If pressing escape closed the window (when no modals were open), then I’d probably use my app launcher (Alfred) to open the window and escape to close it. With this or global hotkey support I wouldn’t mind too much what left click did.

PR #1435, opened 2021-01-28
Add FriendlyName to RLMZone update so correct zone names are displayed to user

### LINKED ISSUE #1432: Zone names with apostrophes or other special characters show blank spaces

Model Name: iPhone 12 Mini
Software Version: 14.3
App version: 2021.1.1 (9)

2021.1.5

**Describe the bug**
When a Zone name continues special characters such as apostrophes or ampersands, these characters do not appear in the zone named displayed in the iOS app, either on the location settings page or in notifications. Instead, the name contains blank spaces and incorrect capitalization. 

**To Reproduce**
Set a zone name containing an apostrophe in HA, then view the Location tab of the App Configuration in the iOS app. The name displayed for the zone will have blank space instead of the expected apostrophe.

**Expected behavior**
Zone name should be displayed consistent with the friendly name from HA. 

**Screenshots**
Defined in `configuration.yaml`: 
Screen Shot 2021-01-27 at 12 27 20 AM

Observed in Developer Tools: 
Screen Shot 2021-01-27 at 12 28 54 AM

Observed in app:
IMG_EAA4F355DED4-1

**Additional context**
It appears that a similar issue was fixed some time back in #308 , but the incorrect display has reappeared. I did a bit of digging, and I suspect that it's because of an issue in `RealmZone.swift`. In the function to return a zone name, [if the self.FriendlyName is nil, the name will be computed from the id](https://github.com/home-assistant/iOS/blob/c6167742f2415658f8dbb7667a136653d2639cdb/Sources/Shared/API/Models/RealmZone.swift#L206-L209), producing a name similar to what I'm observing. And in the update function, it looks like FriendlyName is missing from the [list of properties which are being written to Realm](https://github.com/home-assistant/iOS/blob/c6167742f2415658f8dbb7667a136653d2639cdb/Sources/Shared/API/Models/RealmZone.swift#L53-L60), likely leading to this fallback being used in all cases.

sphanley:
## Summary
This PR adds `FriendlyName` to the list of properties copied over during the `update` function in `RLMZone.swift`, so that the friendly name is shown to the user instead of a generated name based on the ID.

## Screenshots
Compared to the screenshots in #1432, the zone name is now displayed correctly in cases where it contains apostrophes or other characters which don't get included in the ID.

Screen Shot 2021-01-27 at 9 29 16 PM


homeassistant:
Hi @sphanley, 

It seems you haven't yet signed a CLA. Please do so [here](https://home-assistant.io/developers/cla_sign_start/?pr=home-assistant&#x2F;iOS#1435).

Thanks!

zacwest:
Thanks! Good investigation!

PR #1472, opened 2021-02-09
Add setting to control pinch-to-zoom in frontend, defaulted off

zacwest:
Refs home-assistant/frontend#8192.

## Summary
When enabled, pinch to zoom is allowed in the frontend. When disabled, which is the default, it is not allowed. This does not respect what the frontend sends for the values of `minimum-scale`, `maximum-scale` and `user-scalable` and instead overrides them.

## Screenshots
Image

## Any other notes
In home-assistant/frontend#7180, the frontend was changed to enable pinch to zoom in the app by removing the `user-scalable` flag. Browsers ignore this flag, but our UI is meant to feel like an app, not like a webpage. The accessibility part of the argument doesn't mirror iOS accessibility: dynamic type is how the system behaves and it doesn't scale elements like navigation or tab bars. To achieve an iOS-like accessibility experience, we have a setting to adjust only the text size in the frontend.

But we are not in total control over the contents of the users' dashboards or pages. There's layout and design choices that users may make which inherently don't work well on mobile without zooming on non-text elements, including other webapps. I believe this is a rare enough that disabled as a default is acceptable, but also common enough that the preference itself existing is worthwhile.

This works in all versions of iOS that we support, iOS 12 and above.

### DIFF
--- a/Sources/App/Resources/WebSocketBridge.js
+++ b/Sources/App/Resources/WebSocketBridge.js
@@ -2,7 +2,7 @@ const notifyThemeColors = () => {
     function doWait() {
         var colors = {};
 
-        let element = document.createElement('div');
+        const element = document.createElement('div');
         document.body.appendChild(element);
         element.style.display = 'none';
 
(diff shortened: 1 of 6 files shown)


bramkragten:
You know it was reverted right? https://github.com/home-assistant/frontend/pull/8353

TomBrien:
Pinch-to-zoom does work nicely in current iOS app beta (when enabled)

zacwest:
Yeah this setting works either with or without the front end change.

scstraus:
Losing pinch to zoom has made my kiosk tablets borderline useless as I am old with crappy vision and don't wear my glasses all around the house.  Is there any way to enable this on the standard webpage version of the lovelace frontent?

zacwest:
@scstraus Please see the frontend issue linked in the top comment. This change is not directly related to any frontend changes outside of the iOS app.

scstraus:
Yes I read #8192 already, I'm just looking for a solution to the mess it's caused for me with fully kiosk browser where it destroyed zoom. Was hoping someone here might have an idea.

Issue #918, opened 2020-08-16
iOS 14 Widget: Logbook


zacwest:
Show the past few entries in the Logbook, deeplink to the entity in question.

ibru:
Hello @zacwest . I'd like to take a look into this.

I guess in this task I need to:
-  create a iOS 14 widget which will show simple list (Is there any UI designs for it?)
- the list must be driven data from my HA device. How do I get them? Should I use some API request?

zacwest:
The design should probably be modeled after what the frontend version of the logbook looks like. The API is documented on [developers.home-assistant.io](https://developers.home-assistant.io/docs/api/rest).

zacwest:
Rough strokes of what's needed for this:
- Create a new widget in the widget extension. You can use a StaticConfiguration for this one -- it doesn't need Intent-based configuration, so you can skip editing the intentdefinition file.
- Add API calls to the HomeAssistantAPI class which can call the Logbook API referenced above. Use ObjectMapper to create a model representation of the response and it should parse manually.
- In your timeline provider for the widget, invoke the API and stuff the data into the widget.
- Create the widget UI in SwiftUI for the logbook widget.

ibru:
Thanks. This should be sufficient information for now.

Since this will be kind of separated part of the app I could use some design pattern such as MVVM even though it is not used in main app. Do you prefer me to use some of this?
Also I'd use Combine since it works well with SwiftUI. Do you agree?

zacwest:
I believe you can create a VM for the Logbook widget based on the log details to compute what is visible, that should
be okay. Feel free to post model interfaces and we can talk through their features.

Please stick to PromiseKit for any async delivery of details. We may end up skipping combine and going right to coroutines.

ibru:
Created pull request to get Logbook data from API #1481

PR #1481, opened 2021-02-12
Add API request to get logbook entries for current day

ibru:
## Summary
This PR adds method `HomeAssistantAPI.GetLogbook()` used get logbook entries for current day.

Logbook data is needed in order to complete issue #918 "iOS14 Widget: Logbook"

Completing Logbook widget will come in another PR.


Sources/Shared/API/Responses/Logbook.swift
```diff
@@ -0,0 +1,24 @@
+import Foundation
+import ObjectMapper
+
+public class LogbookResponse: Mappable {
```
zacwest: although you followed much of the codebase here, i think you'd be better served by making this conform to `ImmutableMappable` (see e.g. `MobileAppConfigAction`) which allows you to drop all of the not-actually-optional optionals in this class

Sources/Shared/API/Responses/Logbook.swift
```diff
@@ -0,0 +1,24 @@
...
+    public var Domain: String?
```
zacwest: similarly, let's stick with more swifty naming of new things -- variable names should start with a lowercase letter

zacwest (changes requested):

Sources/Shared/API/Responses/Logbook.swift
ibru: I made it conform to `ImmutableMappable `.

However, when doing so I was getting compile errors because there were two identical `request` helper methods, one of them expecting `BaseMappable` and another `ImmutableMappable` (which inherits from `BaseMappable`) so compiler was not sure which one to use.
So I removed the more constraining one for `ImmutableMappable` in 3f76f21fa19fe9bdd9f5381b5b9fd23a1e7d8781

Sources/Shared/API/Responses/Logbook.swift
ibru: Done in 7501e2c5e56b2281923b97370309d4040f00f47b
I was wondering we do we use .NET style for part of the codebase. 

I left the style in `HomeAssistantAPI` as is rest of the class since I think this class would be best to refactor all methods at once.

zacwest: The .NET-style function naming is largely older code; renaming as you go is a fine approach, and keeping that API file consistent is good for now.

Sources/Shared/API/Responses/Logbook.swift
```diff
@@ -0,0 +1,22 @@
+import Foundation
+import ObjectMapper
+
+public struct LogbookResponse: ImmutableMappable {
```
zacwest:
```suggestion
public struct LogbookEntry: ImmutableMappable {
```

Along with renaming the file to match, I think this is a little more clear - `LogbookResponse` would probably be a thing which contains the array, not the items inside the array.

zacwest (changes requested):
lgtm with one minor change to the file name

Sources/Shared/API/Responses/Logbook.swift
ibru: Done in 3a3e23cc374738d5c6ba1124f93f02f55f6ba1f8

zacwest (approved):

PR #1679, opened 2021-06-21
End-to-end encrypted notifications

### LINKED ISSUE #70: Add push notification end to end encryption
Allow users to set a key in their HASS config and on the iOS app which would allow the full payload to be encrypted so that we/AWS/Apple wouldn't be able to see whats in it. Would require a notification service extension and probably migration of a lot of server side code to local. Would also need to expose a bit of metadata still, things like destination device SNS ID, IP of server sending for rate limit purposes, etc.

zacwest:
Fixes #70. Requires https://github.com/home-assistant/core/pull/52048.

- [ ] Pass cleartext commands through, need to wire up the command manager to the register calls. Also still need to make the fcm repo handle them, or make core send them through the legacy path.
- [ ] Avoid erroring the update registration webhook, which doesn't like extra params and so the push_config is going to start erroring for.
- [ ] Tests.
- [ ] Get dependent core PR merged.

## Summary
Allows decrypting notifications in the service extension.

## Any other notes
This relies on changes in core, specifically [my branch](https://github.com/home-assistant/core/compare/dev...zacwest:feat/mobile-app-notify-e2e) of [@robbiet480's branch](https://github.com/home-assistant/core/compare/dev...robbiet480:feat/mobile-app-notify-e2e) which encrypts using the same webhook secret that we already decrypt webhook contents for.

Since we were denied the `com.apple.developer.usernotifications.filtering` entitlement, we need to do the cleartext commands passthrough for things like `clear_notification`.

### DIFF
--- a/Sources/Shared/API/HAAPI.swift
+++ b/Sources/Shared/API/HAAPI.swift
@@ -436,6 +436,14 @@ public class HomeAssistantAPI {
                     "push_url": "https://mobile-apps.home-assistant.io/api/sendPushNotification",
                     "push_token": pushID,
                 ]
+                $0.PushConfig = [
+                    [
+                        "platform": Current.device.systemName(),
+                        "push_token": pushID,
+                        "push_url": "http://localhost:5000/home-assistant-mobile-apps/us-central1/encryptedV1",
+                        "supports_encryption": true,
+                    ],
+                ]
             }
 
             $0.AppIdentifier = Constants.BundleID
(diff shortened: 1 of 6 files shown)


Sources/Shared/API/HAAPI.swift
```diff
@@ -436,6 +436,14 @@ public class HomeAssistantAPI {
                     "push_url": "https://mobile-apps.home-assistant.io/api/sendPushNotification",
                     "push_token": pushID,
                 ]
+                $0.PushConfig = [
+                    [
+                        "platform": Current.device.systemName(),
+                        "push_token": pushID,
+                        "push_url": "http://localhost:5000/home-assistant-mobile-apps/us-central1/encryptedV1",
```
zacwest, line 443: todo

EkilDeew:
Any ideas why this never went out of draft ?
I see the corresponding PR on core was stale as well. 
What happened ?

frenck:
Because there hasn't been any activity on this PR for quite some time now, I've decided to close it for being stale.

Feel free to re-open this PR when you are ready to pick up work on it again 👍

../Frenck

PR #2009, opened 2021-12-18
Disable Bitcode to work around iOS 12 crash

zacwest:
## Summary
Bitcode serves effectively no purpose (it was just for the 32_64 watch migration, and hasn't been used since) but doubly so: starting in ~Xcode 13.2 times, Apple's backport of libConcurrency started crashing on startup but only when they recompile it.


zacwest:
This does not work.

PR #2129, opened 2022-05-12
Publish docker container to ghcr

ludeeus:
## Summary

Moves the publish action from fly.io in preparation of deployment

### DIFF
--- a/.github/workflows/push_deploy.yml
+++ b/.github/workflows/push_deploy.yml
@@ -2,28 +2,35 @@ name: Push Deploy
 
 on:
   workflow_dispatch:
-  push:
-    branches:
-      - master
-    paths:
-      - 'Sources/PushServer/**'
+    inputs:
+      tag:
+        description: 'The tag to give the image'
+        required: true 
+        type: string
 
 concurrency:
   group: '${{ github.workflow }}'
   cancel-in-progress: true
 
-env:
-  FLY_API_TOKEN: ${{ secrets.FLY_API_TOKEN }}
-
 jobs:
   deploy:
-    name: Deploy app
+    name: Deploy image
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
-      - uses: superfly/flyctl-actions@1.1
+
+      - name: Login to GitHub Container Registry
+        uses: docker/login-action@v2.0.0
         with:
-          args: "deploy Sources/PushServer -c app.fly.toml --detach"
-      - uses: superfly/flyctl-actions@1.1
+          registry: ghcr.io
+          username: ${{ github.repository_owner }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Build and push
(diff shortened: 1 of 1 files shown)


.github/workflows/push_deploy.yml
```diff
@@ -1,29 +1,37 @@
...
 on:
-  workflow_dispatch:
-  push:
-    branches:
-      - master
-    paths:
-      - 'Sources/PushServer/**'
+  release:
+    types:
+      - published
```
zacwest: probably not going to be able to do this repo's version numbers to this push server. the downsides of this 'monorepo'. i think it's gotta be these existing -- push to master or workflow dispatch

ludeeus: Why?

ludeeus:
```suggestion
  workflow_dispatch:
    inputs:
      tag:
        description: 'The tag to give the image'
        required: true
        type: string
```

.github/workflows/push_deploy.yml
```diff
@@ -1,29 +1,37 @@
...
+          version=${{ github.ref }}
+          echo "::set-output name=version::$version"
+
+      - name: Build and push
+        uses: docker/build-push-action@v3
         with:
-          args: "deploy Sources/PushServer -c redis.fly.toml --detach"
+          context: ./Sources/PushServer
+          push: true
+          tags: "ghcr.io/${{ github.repository_owner }}/ios-pushserver:${{ steps.version.outputs.version }}"
```
ludeeus:
```suggestion
          tags: "ghcr.io/${{ github.repository_owner }}/ios-pushserver:${{ github.event.inputs.tag }}"
```

.github/workflows/push_deploy.yml
```diff
@@ -1,29 +1,37 @@
...
+          username: ${{ github.repository_owner }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Get version number
+        id: version
+        run: |
+          version=${{ github.ref }}
+          echo "::set-output name=version::$version"
+
+      - name: Build and push
```
ludeeus:
```suggestion
          password: ${{ secrets.GITHUB_TOKEN }}

      - name: Build and push
```

.github/workflows/push_deploy.yml
zacwest: biggest reason is so the "home-assistant-mac.zip" file can be accessible via the latest tag in external links. it also makes it a bit easier to deal with the user-facing nature of the notes too.

ludeeus: This does not change that?

zacwest: It would require we tag more often than we release app updates, and the 'latest' release would no longer be guaranteed to have an attached binary.

ludeeus: It did not have to get additional releases, but lets change it to manual for now then

PR #2131, opened 2022-05-14
Allow trusting otherwise-invalid TLS certificates during onboarding

zacwest:
## Summary
During onboarding, allow trusting an invalid certificate chain for most features of the app.

## Any other notes
Future changes necessary:
- Trusting certificates if an error occurs while the app is running
- Trusting certificates when configuring internal/external URLs


zacwest:
Likely not possible to use AVAssetResourceLoader to enable custom authentication challenges -- shouldWaitForResponseToAuthenticationChallenge is never invoked, and there's many bugs and forum posts of users complaining about this through the years. This eliminates our ability to do just the normal authentication challenge stuff.

Additionally a custom delegate implementation of AVAssetResourceLoader isn't possible as custom URL schemes are only supported for the metadata, not the content, of the HLS playback -- I get a CoreMediaErrorDomain error 12881 which seems to trace back to: custom URL schemes are not supported; it's downloading the content in its own URL Session that doesn't support it.

- https://stackoverflow.com/a/30239876 (someone else trying to do it and failing a few years back)
- https://developer.apple.com/library/archive/technotes/tn2232/_index.html#//apple_ref/doc/uid/DTS40012884-CH1-SECHTTPLIVESTREAMING very old tech note noting this doesn't work; still seems to be the case

zacwest:
Hah, WebKit uses a custom key called `AVURLAssetRequiresCustomURLLoadingKey` which allows the content to be loaded -- at least some of it. https://cs.github.com/WebKit/WebKit/blob/f822d46cdb31d1d3df1915a99c0413acbcb06fd1/Source/WebCore/platform/graphics/avfoundation/objc/MediaPlayerPrivateAVFoundationObjC.mm?q=resourceloaderdelegate#L894

PR #2194, opened 2022-08-30
Added resolutions for Series 7 - 41mm and 45mm watches

d-walsh:


homeassistant:
Hi @dd-w, 

It seems you haven't yet signed a CLA. Please do so [here](https://home-assistant.io/developers/cla_sign_start/?pr=home-assistant&#x2F;iOS#2194).

Thanks!

zacwest:
Updating these numbers (which is correct, I believe) will require we also update all the complication dimension numbers which are keyed on them, see: https://github.com/home-assistant/iOS/blob/ec504fd40f1df3b2ab5e5124838cde3f64ddbe17/Sources/Shared/Common/Extensions/CLKComplication%2BStrings.swift#L1223

PR #2209, opened 2022-09-30
Update README to ensure cocoapods plugin is installed, update Homebrew paths

davidjb:
## Summary

This change to the README adds the installation step required for `cocoapods-acknowledgements` to the readme (where `pod` is installed via Homebrew rather than Bundler). The call to `pod install` would previously fail with:

    [!] Your Podfile requires that the plugin `cocoapods-acknowledgements` be installed. Please install it and try installation again.

This also updates the Homebrew paths used to be dynamically determined, allowing the install instructions to work on Apple Silicon and for any custom Homebrew installation location.

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,12 +18,13 @@ cd iOS
 
 ## install cocoapods via homebrew, use that
 brew install cocoapods
+$(brew --prefix)/opt/ruby/bin/gem install cocoapods-acknowledgements
 pod install --repo-update
 
 ## install ruby via homebrew, use that
 brew install ruby@3.1
-/usr/local/opt/ruby@3.1/bin/bundle install
-/usr/local/opt/ruby@3.1/bin/bundle exec pod install --repo-update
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -18,12 +18,14 @@ cd iOS
 
 ## install cocoapods via homebrew, use that
 brew install cocoapods
+/opt/homebrew/opt/ruby/bin/gem install cocoapods-acknowledgements
```
zacwest: `$(brew --prefix)/opt/ruby/bin/gem` is probably better for this, since it would be automatic

zacwest (changes requested):

README.md
davidjb: Sure thing - I'll update the other lines with this expression as well. Since Homebrew can technically be installed at any path, this'll cover those cases as well.

zacwest (approved):
thanks!

PR #2213, opened 2022-10-01
Add Firefox Focus/Klar as browsers for opening links

davidjb:
## Summary

This adds support to open URLs in Firefox Focus/Klar. Firefox Focus and Klar are the same project
(see https://github.com/mozilla-mobile/focus-ios; and https://support.mozilla.org/en-US/kb/difference-between-firefox-focus-and-firefox-klar), but use different app scheme prefixes on account of the FOCUS name being used by a Mozilla partner in Germany, Austria and Switzerland.

This PR adapts and reuses the existing Firefox URL opener to being able to accept a different scheme for each app, defaulting to the standard `firefox://` prefix for regular Firefox.

## Any other notes

From reading the developer docs, I'm unclear on if/how I should update the `Localizable.strings` files for other languages at this point. If there's something else I need to do for this PR, just let me know 👍 

This has been tested with Firefox Focus on the iOS Simulator. I've tried but as I'm not a subscribed/paid Apple developer, my personal certificate won't allow me to deploy to an actual device, even for debugging (Xcode complains about my personal team not having access to Wifi metadata, amongst a few other missing privileges).

Also, I've done my best with fastline linting/autocorrect but my local setup is having some issue with finding CFPropertyList-3.0.5 and it crashing out at the end. So, it looks okay but again let me know - or CI/GitHub Actions will I'm sure.

### DIFF
--- a/Sources/App/Utilities/OpenInFirefoxControllerSwift.swift
+++ b/Sources/App/Utilities/OpenInFirefoxControllerSwift.swift
@@ -8,12 +8,32 @@ import Foundation
 import UIKit
 
 open class OpenInFirefoxControllerSwift {
-    let firefoxScheme = "firefox:"
-    let basicURL = URL(string: "firefox://")!
+    enum FirefoxType {
+        case regular
+        case focus
+        case klar
+
+        var urlScheme: String {
+            switch self {
+            case .regular:
+                return "firefox"
+            case .focus:
+                return "firefox-focus"
+            case .klar:
+                return "firefox-klar"
+            }
+        }
+    }
+
+    let type: FirefoxType
 
     // This would need to be changed if used from an extension… but you
     // can't open arbitrary URLs from an extension anyway.
     let app = UIApplication.shared
+
+    init(type: FirefoxType = .regular) {
+        self.type = type
+    }
 
     private func encodeByAddingPercentEscapes(_ input: String) -> String {
         NSString(string: input).addingPercentEncoding(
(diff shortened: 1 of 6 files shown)


davidjb:
So fwiw, the issue I was having running any `fastlane` commands was that Ruby commands were erroring out with:

    Could not find CFPropertyList-3.0.5 in any of the sources

and equivalent messages. To solve/workaround the issue, I ran:

    $(brew --prefix)/opt/ruby@3.1/bin/bundle exec fastlane install_plugins

The two `fastlane` plugins it updated seem unrelated, so I can't say precisely why this works, but it did run a step installing plugin dependencies – so I'm guessing that somehow ensured `CFPropertyList` was installed correctly.

Sources/App/Utilities/OpenInFirefoxControllerSwift.swift
```diff
@@ -8,28 +8,31 @@ import Foundation
 import UIKit
 
 open class OpenInFirefoxControllerSwift {
-    let firefoxScheme = "firefox:"
-    let basicURL = URL(string: "firefox://")!
+    var urlScheme: String
```
zacwest: this should probably be an enum, e.g.:

```swift
enum FirefoxType {
  case regular
  case focus
  case klar

  var urlScheme: String {
    switch self { /* … */ }
  }
}
```

zacwest (changes requested):
other languages' localization files are updated automatically (it goes english -> lokalize -> back down to other languages) so you have done the right thing here. just this enum comment and otherwise looks good!

davidjb:
Thanks for the code review @zacwest & insight on localisation. I've added an `enum` as per your suggestion - it's cleaned things up nicely. Tested the adjusted code with Firefox Focus and standard Firefox in the simulator and works well. Let me know if any other changes are needed.

zacwest (approved):
thanks!

PR #2216, opened 2022-10-09
Add option to open links in Private browser tab

davidjb:
## Summary

Adds the option to open links from the companion app within a private browser tab, if the selected browser has this feature (via universal links). The new "Open in Private Tab" UI toggle in settings only shows up when a supported browser is selected.

This option currently only applies to Firefox as it is the only browser to support this (Focus/Klar are private by default and Safari/Chrome don't have a way to open a private or incognito tab at present). That said, the code is written to be extensible should this change later.

Note that if this new option isn't selected, we can't force Firefox to open a link in a "non-private" tab – Firefox iOS [intentionally uses](https://github.com/mozilla-mobile/firefox-ios/blob/main/Client/Application/NavigationRouter.swift#L216-L217) the last mode Firefox was in by default when opening URLs.

## Screenshots

Open in Private tab light screenshot Open in Private tab dark screenshot

### DIFF
--- a/Sources/App/Settings/SettingsDetailViewController.swift
+++ b/Sources/App/Settings/SettingsDetailViewController.swift
@@ -201,6 +201,21 @@ class SettingsDetailViewController: HAFormViewController, TypedRowControllerType
                     prefs.setValue(browserChoice.rawValue, forKey: "openInBrowser")
                 }
 
+                <<< SwitchRow("openInPrivateTab") {
+                    $0.hidden = .function(["openInBrowser"], { form in
+                        if let row = form
+                            .rowBy(tag: "openInBrowser") as? PushRow<OpenInBrowser> {
+                            return row.value?.supportsPrivateTabs == false
+                        } else {
+                            return true
+                        }
+                    })
+                    $0.title = L10n.SettingsDetails.General.OpenInPrivateTab.title
+                    $0.value = prefs.bool(forKey: "openInPrivateTab")
+                }.onChange { row in
+                    prefs.setValue(row.value, forKey: "openInPrivateTab")
+                }
+
                 <<< SwitchRow("confirmBeforeOpeningUrl") {
                     $0.title = L10n.SettingsDetails.Notifications.PromptToOpenUrls.title
                     $0.value = prefs.bool(forKey: "confirmBeforeOpeningUrl")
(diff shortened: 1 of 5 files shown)


Sources/App/Settings/SettingsDetailViewController.swift
```diff
@@ -951,4 +966,8 @@ enum OpenInBrowser: String, CaseIterable {
             return true
         }
     }
+
+    var supportsPrivateTabs: Bool {
+        [.Firefox].contains(self)
```
zacwest: this is fine, though in general you should prefer a `switch` for something like this as it forces you to consider new cases when adding them

Sources/App/Utilities/OpenInFirefoxControllerSwift.swift
```diff
@@ -45,11 +45,12 @@ open class OpenInFirefoxControllerSwift {
         app.canOpenURL(URL(string: "\(type.urlScheme)://")!)
     }
 
-    open func openInFirefox(_ url: URL) {
+    open func openInFirefox(_ url: URL, privateTab: Bool = false) {
```
zacwest, line 48:
```suggestion
    open func openInFirefox(_ url: URL, privateTab: Bool) {
```

since the parameter is always provided, let's drop the default value

zacwest (changes requested):
thanks!

Sources/App/Settings/SettingsDetailViewController.swift
davidjb: Absolutely, makes sense. Changing it to a switch will also allow for more elaborate checks, more than just a hardcoded true/false.

Sources/App/Utilities/OpenInFirefoxControllerSwift.swift
davidjb: The `privateTab` parameter gets provided just for Firefox as derivatives (Firefox Focus and Klar) don't support private tabs and get passed just the `url`.  Is this fine to stay as is?

zacwest: Fair point - I think this can stay with a default parameter

davidjb:
Thanks for the review @zacwest, this should be good to go now.

zacwest (approved):
thanks!

PR #2231, opened 2022-10-20
Enable Realm database compaction on launch

### LINKED ISSUE #2207: iOS app document & data storage size will randomly grow very large
Model Name: iPhone 14 Pro (Noticed this issue on iPhone 11 too)
Software Version: iOS 16.0.1 (Noticed this issue on iOS 15)
App version: 2022.8 

2022.9.6 running on home assistant OS 9.0. Issue was definitely present on previous versions.

**Describe the bug**
Although I cannot pinpoint what causes this to start but once noticed the issue only goes away if the app is deleted and reinstalled although it seems to return. Specifically the "Documents and Data" storage size for the home assistant app grows very large and remains around this size, in my case this will be around 780MB. Deleting the front end cache, location history and event log seems to have no affect. Total size of the zipped logs are around 17MB but nothing relevant that I can see.  

**To Reproduce**
Normal use of the app, nothing specifically relevant in the logs. 

**Expected behavior**
Seems it shouldn't grow so large.

**Screenshots**
Screenshot 2022-09-23 at 1 13 24 AM
datastore

**Additional context**
I was able to use a 3rd party tool to take a backup of the app and extract it from my phone. It seems that of the ~780MB the majority of it is coming from a file called "store.realm" in the dataStore directory (726MB). My understanding is this is potentially created by this https://github.com/home-assistant/iOS/blob/master/Configuration/Podspecs/Realm.podspec.json

I used a tool (https://studio-releases.realm.io/) to open and examine the contents of this file but of the visible data I cannot see how it could come close to 700+MB so it seems potentially there is some corruption?

Not having much experience with ios development I am not sure how relevant this is but this old issue  https://github.com/realm/realm-core/issues/2343 seems to have multiple more recent and potentially relevant links.

zacwest:
Fixes #2207.

## Summary
By default, Realm does not compact its database and [they recommend doing so in production](https://www.mongodb.com/docs/realm/sdk/swift/realm-files/compacting/):

> Every production application should implement a `shouldCompactOnLaunch` callback to periodically reduce the realm file size.

This enables compaction when the database contains more than 50% free space in its on-disk representation. We periodically delete old models at runtime by age, so we expect this to be the case.

### DIFF
--- a/Sources/Shared/Common/Extensions/Realm+Initialization.swift
+++ b/Sources/Shared/Common/Extensions/Realm+Initialization.swift
@@ -212,7 +212,14 @@ public extension Realm {
                     }
                 }
             },
-            deleteRealmIfMigrationNeeded: false
+            deleteRealmIfMigrationNeeded: false,
+            shouldCompactOnLaunch: { realmFileSizeInBytes, usedBytes in
+                // from https://www.mongodb.com/docs/realm/sdk/swift/realm-files/compacting/
+                let maxFileSize = 10 * 1024 * 1024
+                // Check for the realm file size to be greater than the max file size, and the amount of bytes
+                // currently used to be less than 50% of the total realm file size
+                return (realmFileSizeInBytes > maxFileSize) && (Double(usedBytes) / Double(realmFileSizeInBytes)) < 0.5
+            }
         )
 
         do {


jaymunro:
Please also implement a compaction every 6 or 12 hours for ‘kiosk’ devices that are running 24/7 and don’t relaunch. See #2275.

PR #2236, opened 2022-10-21
Swap back to direct pods for Realm/Firebase for stripping reasons

zacwest:
## Summary
Cocoapods' xcframework methodology doesn't strip the generated .framework it makes per architecture/configuration, and modifying that is nontrivial. To make sure we're stripping the binaries, this moves us back to using the official podspecs rather than precompiled xcframeworks.


zacwest:
This first approach did not work because Cocoapods copies the binaries from the xcframework to a destination manually and not as part of a target.

PR #2238, opened 2022-10-22
Move alternate app icons into asset catalog

zacwest:


zacwest:
This errors at archive time with "ERROR: Identical key for two renditions" because we are combining single-size and all-size app icons at the same time. We want to specify different icons for certain sizes, so moving all to "single size" won't work. Will probably need to move all to "all sizes" which is annoying.

PR #2392, opened 2023-08-09
Navigate to default server when deeplink to the app passing "?server=default"

bgoncal:
## Summary
Navigate to default server when deeplink to the app passing "?server=default".
Currently when you use deep link to navigate to one lovelace page and you have multiple HA servers in your iOS App it always asks which server you want to use.
With this change you are able to default to your current server by using a query item "?server=default"

https://github.com/home-assistant/iOS/assets/5808343/c29ac74e-8ea4-4d5a-93ec-53ef7c64ced6

### DIFF
--- a/Sources/App/WebView/WebViewWindowController.swift
+++ b/Sources/App/WebView/WebViewWindowController.swift
@@ -216,8 +216,10 @@ class WebViewWindowController {
                 open(from: from, server: first, urlString: openUrlRaw, skipConfirm: skipConfirm)
             } else {
                 if let selectedServer = servers.first(where: { server in
-                    let sanitizedServerName = server.info.name.lowercased().trimmingCharacters(in: .whitespacesAndNewlines)
-                    let sanitizedSelectedServerName = serverName?.lowercased().trimmingCharacters(in: .whitespacesAndNewlines)
+                    let sanitizedServerName = server.info.name.lowercased()
+                        .trimmingCharacters(in: .whitespacesAndNewlines)
+                    let sanitizedSelectedServerName = serverName?.lowercased()
+                        .trimmingCharacters(in: .whitespacesAndNewlines)
                     return sanitizedServerName == sanitizedSelectedServerName
                 }) {
                     open(from: from, server: selectedServer, urlString: openUrlRaw, skipConfirm: skipConfirm)
--- a/Sources/App/WebView/WebViewWindowController.swift
+++ b/Sources/App/WebView/WebViewWindowController.swift
@@ -217,9 +217,9 @@ class WebViewWindowController {
             } else {
                 if let selectedServer = servers.first(where: { server in
                     let sanitizedServerName = server.info.name.lowercased()
-                        .trimmingCharacters(in: .whitespacesAndNewlines)
+                        .addingPercentEncoding(withAllowedCharacters: .urlQueryAllowed)
                     let sanitizedSelectedServerName = serverName?.lowercased()
-                        .trimmingCharacters(in: .whitespacesAndNewlines)
+                        .addingPercentEncoding(withAllowedCharacters: .urlQueryAllowed)
                     return sanitizedServerName == sanitizedSelectedServerName
                 }) {
                     open(from: from, server: selectedServer, urlString: openUrlRaw, skipConfirm: skipConfirm)
--- a/Sources/App/WebView/WebViewWindowController.swift
+++ b/Sources/App/WebView/WebViewWindowController.swift
@@ -216,9 +216,7 @@ class WebViewWindowController {
                 open(from: from, server: first, urlString: openUrlRaw, skipConfirm: skipConfirm)
             } else {
                 if let selectedServer = servers.first(where: { server in
-                    let sanitizedServerName = server.info.name.lowercased()
-                        .addingPercentEncoding(withAllowedCharacters: .urlQueryAllowed)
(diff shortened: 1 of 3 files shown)


robbiet480 (approved):

cpvbruno:
Im gonna write down unit tests asap, but it may take a while because these classes can't be easily mocked so a refactor may be needed unless I make a UI test

cpvbruno:
@robbiet480 is the code coverage a blocker to merge this PR? From what I can see these classes need a relevant amount of refactoring to be able to be mocked and tested

zacwest:
Don't worry about the code coverage. The UI side doesn't have much. I'll merge this when I am able to give it a test. I may also slip in a simple text match against server names for other string values.

cpvbruno:
Thanks @zacwest, also I was struggling yesterday to open deeplinks in the simulator, do I need to use something different than "homeassistant://" when testing on simulator?

zacwest:
The URLScheme for debug builds is different. homeassistant-dev://…

cpvbruno:
I have updated a PR with a small fix and also added a video in the PR description showing up the behaviour.
What's not shown there is an alert that home assistant has asking if it should open the Dee-link that time or always, I tapped always before recording this.

cpvbruno:
> Don't worry about the code coverage. The UI side doesn't have much. I'll merge this when I am able to give it a test. I may also slip in a simple text match against server names for other string values.

I updated the PR once more, I added the option to point the server you want to use by passing it's name instead of "default", is that what you meant?

cpvbruno:
Hey @zacwest did you had time to give it a try? Let me know if I can help/improve something

Sources/App/WebView/WebViewWindowController.swift
```diff
@@ -202,9 +202,31 @@ class WebViewWindowController {
...
+        if let first = servers.first, Current.servers.all.count == 1 || serverName != nil {
+            if serverName == "default" || serverName == nil {
+                open(from: from, server: first, urlString: openUrlRaw, skipConfirm: skipConfirm)
+            } else {
+                if let selectedServer = servers.first(where: { server in
+                    let sanitizedServerName = server.info.name.lowercased()
+                        .addingPercentEncoding(withAllowedCharacters: .urlQueryAllowed)
+                    let sanitizedSelectedServerName = serverName?.lowercased()
+                        .addingPercentEncoding(withAllowedCharacters: .urlQueryAllowed)
+                    return sanitizedServerName == sanitizedSelectedServerName
```
zacwest: can we fuzzy match? for example, if i have:

1. `zac's house`
2. `the shed`

can we match with `?server=house` and `?server=shed`?

Sources/App/WebView/WebViewWindowController.swift
```diff
@@ -202,9 +202,31 @@ class WebViewWindowController {
...
+        let serverName = queryParameters?.first(where: { $0.name == "server" })?.value
+        let servers = Current.servers.all
+
+        if let first = servers.first, Current.servers.all.count == 1 || serverName != nil {
+            if serverName == "default" || serverName == nil {
+                open(from: from, server: first, urlString: openUrlRaw, skipConfirm: skipConfirm)
+            } else {
+                if let selectedServer = servers.first(where: { server in
+                    let sanitizedServerName = server.info.name.lowercased()
+                        .addingPercentEncoding(withAllowedCharacters: .urlQueryAllowed)
```
zacwest: doesn't the URLQueryItem version of this strip url encoding?

zacwest (changes requested):

Sources/App/WebView/WebViewWindowController.swift
bgoncal: It's doable but should we? I think this may lead to name conflict at some point

Sources/App/WebView/WebViewWindowController.swift
cpvbruno: I don't remember why I did that but I remember I tested some scenarios that the browser allowed but then the comparison would fail... something related to non ASCII chars.
I'll test again and get back to you

bgoncal: Good catch, I have updated the PR

Sources/App/WebView/WebViewWindowController.swift
bgoncal: Like If I have 2 server "House" and "Parent's House", using fuzzy match on "House" will conflict

zacwest: hm, i think something like `server=house` would match `house` over `parent's house` -- maybe the 'closer' the better. maybe that's good for a second round at this.

Sources/App/WebView/WebViewWindowController.swift
zacwest: if the value in the URLQueryItem is decoded already, wouldn't this make something 'weird' not match?

zacwest: yeah so this introduces an issue with spaces here. `?server=dev%20local` looks like:

```
sanitizedServerName: "dev%20local"
serverName: "dev local"
```

i dont' think we need to do any url in this line - we're on both sides of the "not encoded"

bgoncal: You're correct "addingPercentEncoding" was irrelevant there, I have removed it

zacwest (approved):

PR #2398, opened 2023-08-17
Biometrics lock (Lock app with FaceID/TouchID/iOS Password)

bgoncal:
## Summary
Concept for locking the HA app using iOS biometrics. (FaceID wasn't working on my simulator, thats why the video shows password instead).

Pending:
- Localised copy
- Better error logging
- Remove print statements


Sources/App/WebView/LocalAuthentication/AuthenticationService.swift
```diff
@@ -0,0 +1,36 @@
...
+            authenticate(policy: .deviceOwnerAuthenticationWithBiometrics)
+        } else if context.canEvaluatePolicy(.deviceOwnerAuthentication, error: &error) {
+            authenticate(policy: .deviceOwnerAuthentication)
+        } else {
+            print("Failed to authenticate")
+        }
+    }
+
+    private func authenticate(policy: LAPolicy) {
+        context.evaluatePolicy(policy, localizedReason: "Authentication required") { approved, error in
```
zacwest: the hard part here is also blocking the ui while this should be happening, and e.g. in the app switcher. how would you imagine this to be implemented?

Sources/App/WebView/WebViewController.swift
```diff
@@ -24,6 +24,7 @@ class WebViewController: UIViewController, WKNavigationDelegate, WKUIDelegate, U
 
     var keepAliveTimer: Timer?
     private var initialURL: URL?
+    private let authenticationService: AuthenticationServiceProtocol
```
zacwest: DI using `Current` (see https://www.pointfree.co/blog/posts/21-how-to-control-the-world)

Sources/App/Settings/SettingsRootDataSource.swift
```diff
@@ -95,6 +97,16 @@ enum SettingsRootDataSource {
         }
     }
 
+    private static func security() -> SettingsButtonRow {
+        SettingsButtonRow {
+            $0.title = "Security"
```
zacwest: you can wrap in `NSLocalizedString` as a reminder to add it to the L10n/swiftgen stuff, makes it a little easier to avoid missing things

zacwest (changes requested):

Sources/App/WebView/LocalAuthentication/AuthenticationService.swift
bgoncal: Usually most apps blur the UI using and overlay UIView, what do you think? I'd go for that, although at least in the simulator I wasn't able to dismiss iOS biometric/password screen (which is visible in the video), so maybe it is already doing that.

Sources/App/WebView/WebViewController.swift
bgoncal: Good one! I was looking for which DI the project used, thanks!

Sources/App/Settings/SettingsRootDataSource.swift
bgoncal: How does the translation flow works? Like... I can input English and Portuguese, does the community fills the rest?

Sources/App/WebView/LocalAuthentication/AuthenticationService.swift
bgoncal: Something like this, I just added to the PR, but Im not 100% sure the App Switcher will show this overlay if the user has already authenticated.
Simulator Screenshot - iPhone SE (3rd generation) - 2023-08-21 at 22 44 06

Sources/App/WebView/WebViewController.swift
bgoncal: And of course that was my first reaction "Isn’t Current a singleton? Aren’t singletons evil?" haha like they have in the FAQ

Sources/App/Settings/SettingsRootDataSource.swift
bgoncal: I have wrapped all strings I added for now

Sources/App/Resources/en.lproj/Localizable.strings
```diff
@@ -410,6 +410,9 @@ Home Assistant is free and open source home automation software with a focus on
 "settings_details.general.visibility.options.dock_and_menu_bar" = "Dock and Menu Bar";
 "settings_details.general.visibility.options.menu_bar" = "Menu Bar";
 "settings_details.general.visibility.title" = "Show App In…";
+"settings_details.general.security.title" = "Security";
+"settings_details.general.security.action" = "Enable biometric lock";
+"settings_details.general.security.footer" = "You will be required to authenticate with biometrics everytime you open the app.";
```
zacwest: `every time`

Sources/App/WebView/WebViewController.swift
```diff
@@ -24,6 +24,10 @@ class WebViewController: UIViewController, WKNavigationDelegate, WKUIDelegate, U
 
     var keepAliveTimer: Timer?
     private var initialURL: URL?
+    private let biometricOverlay: UIView = {
```
zacwest: let's move this to either the WebViewWindowController or something on the Scene level. if it's not related to the function of the WebView, it should go elsewhere

Sources/App/WebView/WebViewController.swift
```diff
@@ -616,6 +654,43 @@ class WebViewController: UIViewController, WKNavigationDelegate, WKUIDelegate, U
...
+            DispatchQueue.main.async { [weak self] in
+                self?.biometricOverlay.removeFromSuperview()
+            }
+        }
+    }
+
+    private func addBiometricOverlayProtection() {
+        DispatchQueue.main.async { [weak self] in
+            guard let self else { return }
+            self.biometricOverlay.removeFromSuperview()
```
zacwest: make sure you set `accessibilityViewIsModal` and double check that VoiceOver can't get past this overlay

Sources/Shared/LocalAuthentication/AuthenticationService.swift
```diff
@@ -0,0 +1,49 @@
...
+            authenticate(policy: .deviceOwnerAuthenticationWithBiometrics)
+        } else if context.canEvaluatePolicy(.deviceOwnerAuthentication, error: &error) {
+            authenticate(policy: .deviceOwnerAuthentication)
+        } else {
+            Current.Log
+                .error(
+                    [
+                        "Failed to authenticate with biometrics": "User context can't evaluate policy for device ownership",
+                    ]
+                )
```
zacwest: what should we tell the user if they fail to do this? should we have on-screen ui that says an error?

Sources/Shared/LocalAuthentication/AuthenticationService.swift
```diff
@@ -0,0 +1,49 @@
...
+        }
+    }
+
+    func invalidate() {
+        context.invalidate()
+    }
+
+    private func authenticate(policy: LAPolicy) {
+        context.evaluatePolicy(policy, localizedReason: "Authentication required") { [weak self] authorized, error in
```
zacwest: needs to be translated

Sources/Shared/LocalAuthentication/AuthenticationService.swift
```diff
@@ -0,0 +1,49 @@
...
+    func authenticate() {
+        var error: NSError?
+        if context.canEvaluatePolicy(.deviceOwnerAuthenticationWithBiometrics, error: &error) {
+            authenticate(policy: .deviceOwnerAuthenticationWithBiometrics)
+        } else if context.canEvaluatePolicy(.deviceOwnerAuthentication, error: &error) {
+            authenticate(policy: .deviceOwnerAuthentication)
+        } else {
+            Current.Log
+                .error(
+                    [
```
zacwest: is this a dictionary for any particular reason?

zacwest (changes requested):

Sources/App/Settings/SettingsRootDataSource.swift
```diff
@@ -95,6 +97,16 @@ enum SettingsRootDataSource {
         }
     }
 
+    private static func security() -> SettingsButtonRow {
+        SettingsButtonRow {
```
zacwest, line 101: how does this work on catalyst? should we hide it or make it available -- if the latter, how do we deal with e.g. the settings scene?

bgoncal: In catalyst it was a bit trickier, because of the "enter background/foreground" thats happens all the time even with the touchID dialog, I treated it with an extra logic to only reauthenticate when the user has reply enter background.

When the user open settings scene, the main web view scene locks again until the user unlocks it.

https://github.com/home-assistant/iOS/assets/5808343/7fdb056c-5339-476a-be25-f55dbec457cd

https://github.com/home-assistant/iOS/assets/5808343/6a0292af-024d-40d2-8899-059da3c3f62e

Sources/Shared/LocalAuthentication/AuthenticationService.swift
bgoncal: Nope, removed

Sources/Shared/LocalAuthentication/AuthenticationService.swift
bgoncal: If the user can't authenticate neither with biometrics or password then the user is locked out, I would say this is very edge case since it's credentials are the same as the iOS/macOS device. Any suggestion?

Sources/App/WebView/WebViewController.swift
bgoncal: I replaced the overlay by a full view controller

Sources/Shared/LocalAuthentication/BiometricsAuthenticationViewController.swift
```diff
@@ -0,0 +1,87 @@
...
+        view.addSubview(stackView)
+        stackView.translatesAutoresizingMaskIntoConstraints = false
+        NSLayoutConstraint.activate([
+            stackView.leadingAnchor.constraint(equalTo: view.leadingAnchor),
+            stackView.trailingAnchor.constraint(equalTo: view.trailingAnchor),
+            stackView.centerYAnchor.constraint(equalTo: view.centerYAnchor),
+        ])
+
+        view.layoutIfNeeded()
+        unlockButton.addGestureRecognizer(UITapGestureRecognizer(target: self, action: #selector(didTapUnlock)))
```
zacwest: let's use contentEdgeInsets on the button to get the look you're going for, so you don't have to do the gesture recognizer here.

if you do want to do the gesture recognizer, please set an `accessibilityLabel` and appropriate `accessibilityTraits` so this is exposed as a button

Sources/Shared/LocalAuthentication/BiometricsAuthenticationViewController.swift
```diff
@@ -0,0 +1,87 @@
...
+
+        containerView.addSubview(button)
+        button.translatesAutoresizingMaskIntoConstraints = false
+        NSLayoutConstraint.activate([
+            button.topAnchor.constraint(equalTo: containerView.topAnchor, constant: 10),
+            button.bottomAnchor.constraint(equalTo: containerView.bottomAnchor, constant: -10),
+            button.leadingAnchor.constraint(equalTo: containerView.leadingAnchor, constant: 16),
+            button.trailingAnchor.constraint(equalTo: containerView.trailingAnchor, constant: -16),
+        ])
+        containerView.backgroundColor = .homeAssistant
```
zacwest: try swapping light/dark mode here, i'm guessing things may go wonky. can you look at `Style` and the onboarding button configs there? that may be a good thing to steal/use

Sources/Shared/LocalAuthentication/BiometricsAuthenticationViewController.swift
```diff
@@ -0,0 +1,87 @@
...
+        imageView.frame = CGRect(x: 0, y: 0, width: 256, height: 256)
+        imageView.contentMode = .scaleAspectFit
+        return imageView
+    }()
+
+    private let unlockButton: UIView = {
+        let containerView = UIView()
+        let button = UIButton()
+        button.setTitle(L10n.SettingsDetails.General.Security.unlockButton, for: .normal)
+        button.titleLabel?.font = UIFont.preferredFont(forTextStyle: .headline)
```
zacwest: how does this end up getting its title color set? i don't see relevant code but maybe i missed it

Sources/Shared/LocalAuthentication/BiometricsAuthenticationViewController.swift
```diff
@@ -0,0 +1,87 @@
...
+    }
+
+    public func updateUnlockButtonVisibility(visible: Bool) {
+        DispatchQueue.main.async { [weak self] in
+            self?.unlockButton.isHidden = !visible
+        }
+    }
+
+    private func setupUI() {
+        if #available(iOSApplicationExtension 13.0, *) {
```
zacwest: `iOS 13.*` should be good enough

Sources/WatchApp/Assets.xcassets/WatchIcon.appiconset/Contents.json
zacwest, line 1: accidental changes here?

Sources/App/Settings/Security/SecurityViewController.swift
```diff
@@ -0,0 +1,28 @@
...
+
+        title = L10n.SettingsDetails.General.Security.title
+
+        form
+            +++ Section()
+            <<< SwitchRow("switch", {
+                $0.title = L10n.SettingsDetails.General.Security.action
+                $0.value = Current.settingsStore.biometricsRequired
+                $0.onChange { row in
+                    Current.settingsStore.biometricsRequired = row.value ?? false
```
zacwest, line 21: since settings may be open when locked, can you make it require authentication to disable?

Sources/App/WebView/WebViewWindowController.swift
```diff
@@ -469,3 +471,48 @@ extension WebViewWindowController: OnboardingStateObserver {
         }
     }
 }
+
+// MARK: - Biometrics lock
+
+extension WebViewWindowController {
+    private func listenForBiometricsLockRelatedEvents() {
```
zacwest, line 478: what happens if the user gets signed out (e.g. remotely,  due to token) while this is happening?

Sources/Shared/LocalAuthentication/BiometricsAuthenticationService.swift
```diff
@@ -0,0 +1,142 @@
...
+    @objc private func allowReauthenticate() {
+        shouldAllowMacToAuthenticate = true
+    }
+    #endif
+
+    private func authenticate() {
+        var error: NSError?
+        context?.invalidate()
+        context = LAContext()
+        if context?.canEvaluatePolicy(.deviceOwnerAuthenticationWithBiometrics, error: &error) ?? false {
```
zacwest: should we allow watch too?

Sources/Shared/LocalAuthentication/BiometricsAuthenticationService.swift
```diff
@@ -0,0 +1,142 @@
...
+
+public protocol BiometricsAuthenticationServiceProtocol {
+    func checkForBiometrics(controller: UIViewController)
+    func protectAppIfNeeded(controller: UIViewController, completion: (() -> Void)?)
+}
+
+class BiometricsAuthenticationService: BiometricsAuthenticationServiceProtocol {
+    private var context: LAContext?
+
+    private var overlayViewController: BiometricsAuthenticationViewController?
```
zacwest: can we talk to this without knowing it's a UIViewController?

Sources/Shared/LocalAuthentication/BiometricsAuthenticationService.swift
```diff
@@ -0,0 +1,142 @@
...
+            completion()
+            return
+        }
+        overlayViewController = BiometricsAuthenticationViewController()
+        overlayViewController?.modalPresentationStyle = .overCurrentContext
+        overlayViewController?.delegate = self
+        guard let overlayViewController = overlayViewController else { return }
+
+        DispatchQueue.main.async {
+            controller.present(overlayViewController, animated: false) {
```
zacwest: what happens if a modal is already being presented?

i think we may need to do something more like:

1. insert this into the window at the top
2. do the appearance manually (beginTransition/endTransition)
3. make sure it stays on top

zacwest (changes requested):

Sources/Shared/LocalAuthentication/BiometricsAuthenticationViewController.swift
bgoncal: I removed the UIKit implementation and added a SwiftUI view instead, using the design system button created in the other PR (HAButton)

Sources/Shared/LocalAuthentication/BiometricsAuthenticationViewController.swift
bgoncal: UIKit implementation removed

Sources/App/Settings/Security/SecurityViewController.swift
bgoncal: I removed catalyst implementation and kept it just for iOS for now to keep it simple and iterate afterwards. 
So in iOS the user has to unlock the app before going to settings.

Sources/App/WebView/WebViewWindowController.swift
bgoncal: How can I easily reproduce that? I think the overlay will stay there, but it's nice to verify

Sources/Shared/LocalAuthentication/BiometricsAuthenticationService.swift
bgoncal: I've never seen this happening inside apps, only unlocking iPhone itself, can you point me out to the policy?

Sources/Shared/LocalAuthentication/BiometricsAuthenticationService.swift
bgoncal: I made it agnostic to controller class now, but still a UIViewController reference. Are you concerned about testability?

bgoncal: Done, now it replaces the rootViewController of the window with the lock screen and then replaces back after authenticating

Sources/WatchApp/Assets.xcassets/WatchIcon.appiconset/Contents.json
bgoncal: Reverted

Sources/App/Settings/SettingsViewController.swift
```diff
@@ -102,11 +102,19 @@ class SettingsViewController: HAFormViewController {
         }
 
         if contentSections.contains(.general) {
+            #if targetEnvironment(macCatalyst)
+            form +++ Section()
```
zacwest: a few things here:

you can do a `let section = Section()` and append to it then `form +++ section` if you wanna avoid duplicated code here

you can also use the section's visibility modifier to hide on catalyst, e.g.

```
Section {
  $0.hidden = .isCatalyst
}
```

you may also find `Current.isCatalyst` easier than target defines -- less weird code paths

Sources/App/WebView/WebViewWindowController.swift
zacwest: go into your user profile & delete the token for your device in your profile. you can do it from the same machine or other.

if the overlay stays, it may be worth killing it off since there's nothing it's protecting (if it's the last account) and may block signing into something new. there's no way to get to settings during onboarding.

Sources/Shared/DesignSystem/Constants/HASpacing.swift
```diff
@@ -0,0 +1,15 @@
+import Foundation
+
+public enum Spacing {
+    public static let one: CGFloat = 8
+    public static let one_and_a_half: CGFloat = 12
```
zacwest: camelCase?

Sources/Shared/LocalAuthentication/BiometricsAuthenticationService.swift
zacwest: `kLAPolicyDeviceOwnerAuthenticationWithBiometricsOrWatch`

HomeAssistant.xcodeproj/project.pbxproj
```diff
@@ -4554,8 +4583,6 @@
 				"pt-BR",
 				ml,
 				hu,
-				he,
-				et,
```
zacwest: accidental changes in here?

HomeAssistant.xcodeproj/project.pbxproj
```diff
@@ -6326,8 +6360,6 @@
 				11B154A327ABAADA00462185 /* pt-BR */,
 				1128FF38297E5F7D00BAAFD9 /* ml */,
 				11EFB44C29D89FD100CE4B05 /* hu */,
-				42DD84332B15DC2F00936F16 /* he */,
-				42DD84372B15DC3F00936F16 /* et */,
```
zacwest, line 6330: yeah i think the xcodeproj merge went bad

Sources/App/Resources/Info.plist
zacwest, line 1: anything happening in here?

zacwest (changes requested):

Sources/Shared/LocalAuthentication/BiometricsAuthenticationService.swift
bgoncal: This is available only for Mac, which this PR won't support yet.

Sources/App/Resources/Info.plist
bgoncal: `NSFaceIDUsageDescription`

Sources/App/WebView/WebViewWindowController.swift
bgoncal: Tested. Indeed the "lock screen" was still being shown even with no servers. So I updated it in a way that the user needs to authenticate one last time after the token expires, and then the biometric lock will be disabled and only reenabled if the user sets up a new server and manually enables the toggle.

HomeAssistant.xcodeproj/project.pbxproj
bgoncal: Indeed this merge was problematic, I've readded those languages, I think no other file is missing anymore

Sources/App/Resources/et.lproj/Localizable.strings
```diff
@@ -769,3 +775,8 @@ Home Assistant is free and open source home automation software with a focus on
 "widgets.open_page.not_configured" = "No Pages Available";
 "widgets.open_page.title" = "Open Page";
 "yes_label" = "Yes";
+"thread.credentials.network_name_title" = "Network Name";
+"thread.credentials.network_key_title" = "Network Key";
+"thread.credentials.border_agent_id_title" = "Border Agent ID";
+"thread.credentials.share_credentials_button_title" = "Share credential with Home Assistant";
+"thread.credentials.screen_title" = "Thread Credentials";
```
zacwest, line 782: bad merge?

Sources/Shared/LocalAuthentication/BiometricsAuthenticationService.swift
```diff
@@ -0,0 +1,109 @@
...
+    private func addBiometricOverlayProtection(controller: UIViewController, completion: @escaping () -> Void) {
+        guard #available(iOS 13.0, *) else { return }
+
+        let overlayController = UIHostingController(rootView: BiometricsView.build(delegate: self))
+        overlayController.modalPresentationStyle = .overCurrentContext
+
+        guard let window = controller.view.window,
+              previousRootViewController == nil else { return }
+
+        self.window = window
```
zacwest, line 97: what happens if there are multiple windows open? e.g. multi-scene support on iPad.

i think you probably want to either collate all the windows during the lock phase, or make this chain off the WebViewWindowController itself, but synchronize the unlock status

zacwest (changes requested):
i think things seem good, but i am a little worried about the 'protect' part and how it works with multiple windows happening

Sources/Shared/LocalAuthentication/BiometricsAuthenticationService.swift
bgoncal: I think I will start over with this implementation, I liked your idea of sync status. Another question... should we allow shortcut actions if the app is locked?

zacwest: I think it becomes really hard to do locking if we disallow extensions to do work. Most of them do not support being able to prompt the user for permission.

We could either:
- tell the user it won't lock extensions
- disallow extensions entirely if lock is enabled (where we cannot prompt)
- allow the user to choose which extensions should be allowed

Maybe we could add more control in app but I think the first one sounds easiest for now.

bgoncal:
Closing PR, I will start over with a better foundation for locking the app

PR #2435, opened 2023-10-25
Less Array

RomanPodymov:
Hello.
Thank you for home-assistant/iOS.

## Summary

Using less `Array.init`.

### DIFF
--- a/Sources/App/Settings/Eureka/RealmSection.swift
+++ b/Sources/App/Settings/Eureka/RealmSection.swift
@@ -93,7 +93,7 @@ public final class RealmSection<ObjectType: Object>: Section {
             }.flatMap { $0 }
 
             // note: eureka has a bug if we don't wrap this in Array where it duplicates
-            append(contentsOf: Array(appendingRows))
+            append(contentsOf: appendingRows)
         }
 
         didUpdate(self, collection)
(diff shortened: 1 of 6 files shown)


robbiet480:
Why?

Sources/App/Settings/Eureka/RealmSection.swift
```diff
@@ -93,7 +93,7 @@ public final class RealmSection<ObjectType: Object>: Section {
             }.flatMap { $0 }
 
             // note: eureka has a bug if we don't wrap this in Array where it duplicates
```
zacwest, line 95: Did you test this?

zacwest (changes requested):

Sources/App/Settings/Eureka/RealmSection.swift
bgoncal: @RomanPodymov I see your reaction to Zac's comment, could you confirm you tested the part of the code you have updated? I understand it looks so simple that it is not required but sometimes we break things with small changes like this, and I really appreciate the "scout" on HA code, thank you, so please let me know otherwise I will close this PR for now.

RomanPodymov: Hello @bgoncal 
Let's close it for now.

PR #2437, opened 2023-11-01
Toggle scenes/actions for watchOS App all at once

bgoncal:
## Summary
When you open your watchOS App for the first time you may encounter several scenes already active to be used in the watch App (specially if you have Philips Hue in HA), this change allows you to unselect all scenes at once (hide from watch app) if you want.

## Screenshots
https://github.com/home-assistant/iOS/assets/5808343/e2f0f9fb-704d-4240-a817-4cf7318269b3

### DIFF
--- a/Sources/App/Settings/SettingsDetailViewController.swift
+++ b/Sources/App/Settings/SettingsDetailViewController.swift
@@ -22,6 +22,7 @@ class SettingsDetailViewController: HAFormViewController, TypedRowControllerType
     private var notificationTokens: [NotificationToken] = []
     private var notificationCenterTokens: [AnyObject] = []
     private var reorderingRows: [String: BaseRow] = [:]
+    private var scenesUpdateObserver: NotificationToken?
 
     deinit {
         notificationCenterTokens.forEach(NotificationCenter.default.removeObserver)
@@ -443,6 +444,29 @@ class SettingsDetailViewController: HAFormViewController, TypedRowControllerType
             )
 
             let scenes = realm.objects(RLMScene.self).sorted(byKeyPath: RLMScene.positionKeyPath)
+
+            let toggleAllSwitch = SwitchRow()
+            toggleAllSwitch.title = L10n.SettingsDetails.Actions.Scenes.selectAll
+
+            scenesUpdateObserver = scenes.observe { _ in
+                // Access UISwitch directly to set state to avoid triggering "on value change"
+                toggleAllSwitch.cell.switchControl.setOn(
+                    scenes.filter(\.actionEnabled).count == scenes.count,
+                    animated: true
+                )
+            }
+
+            toggleAllSwitch.onChange { [weak self] row in
+                guard let self = self,
+                      let value = row.value else { return }
+                self.realm.reentrantWrite {
+                    scenes.forEach { scene in
+                        scene.actionEnabled = value
+                    }
+                }
+            }
+            form +++ toggleAllSwitch
+
             form +++ RealmSection<RLMScene>(
                 header: L10n.SettingsDetails.Actions.Scenes.title,
                 footer: L10n.SettingsDetails.Actions.Scenes.footer,
@@ -453,26 +477,7 @@ class SettingsDetailViewController: HAFormViewController, TypedRowControllerType
                         $0.disabled = true
                     },
                 ], getter: {
-                    var baseRows: [BaseRow] = []
-                    if $0 == scenes.first {
-                        let toggleAllSwitch = SwitchRow()
-                        toggleAllSwitch.title = L10n.SettingsDetails.Actions.Scenes.selectAll
-                        toggleAllSwitch.value = scenes.filter(\.actionEnabled).count == scenes.count
-                        toggleAllSwitch.onChange { [weak self] row in
-                            guard let self = self,
-                                  let value = row.value else { return }
-                            self.realm.reentrantWrite {
-                                scenes.forEach { scene in
-                                    scene.actionEnabled = value
-                                }
-                            }
-                        }
-                        baseRows = [toggleAllSwitch]
-                    }
-
-                    baseRows += Self.getSceneRows($0)
-
-                    return baseRows
+                    Self.getSceneRows($0)
                 }
             )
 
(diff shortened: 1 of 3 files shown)


zacwest:
The real crux of the issue is the startup routine for HA causes the entire scene domain to have nothing in it, which makes the app remove all its local copies. Once it later populates the scenes, it recreates them and loses the disabled state. A good fix for this probably would involve not listening to state changes until core finishes loading all the way.

I think some integrations may also work by deleting all then restoring them, so moving enabled state out of Realm might also be required.

Sources/App/Settings/SettingsDetailViewController.swift
```diff
@@ -454,7 +453,29 @@ class SettingsDetailViewController: HAFormViewController, TypedRowControllerType
                         $0.disabled = true
                     },
                 ], getter: {
-                    Self.getSceneRows($0)
+                    var baseRows: [BaseRow] = []
+                    if $0 == scenes.first {
+                        let toggleAllSwitch = SwitchRow()
+                        toggleAllSwitch.title = NSLocalizedString(
+                            "Select All",
```
zacwest: We don't run genstrings at all, you need to modify Localizable.strings and use the Swiftgen-generated L10n module to access them with compile time safety.

Sources/App/Settings/SettingsDetailViewController.swift
```diff
@@ -454,7 +453,29 @@ class SettingsDetailViewController: HAFormViewController, TypedRowControllerType
...
+                            comment: "Toggle to select/unselect all actions that are displayed in watchOS app"
+                        )
+                        toggleAllSwitch.value = scenes.filter(\.actionEnabled).count == scenes.count
+                        toggleAllSwitch.onChange { [weak self] row in
+                            guard let self = self,
+                                  let value = row.value else { return }
+                            self.realm.beginWrite()
```
zacwest: Use the reentrantWrite wrapper

zacwest (changes requested):
Good problem to tackle

Sources/App/Settings/SettingsDetailViewController.swift
bgoncal: Got it, I've added a new English string for it then

bgoncal:
> The real crux of the issue is the startup routine for HA causes the entire scene domain to have nothing in it, which makes the app remove all its local copies. Once it later populates the scenes, it recreates them and loses the disabled state. A good fix for this probably would involve not listening to state changes until core finishes loading all the way.
> 
> I think some integrations may also work by deleting all then restoring them, so moving enabled state out of Realm might also be required.

Indeed some clean up needs to be done in the startup flow, I'll put it on my to-do list to look into it since it may require more time for me that's not familiar with it yet.
For now, I believe this "Select all" toggle is a quick win anyway, what do you think?

Sources/App/Settings/SettingsDetailViewController.swift
```diff
@@ -454,7 +453,26 @@ class SettingsDetailViewController: HAFormViewController, TypedRowControllerType
                         $0.disabled = true
                     },
                 ], getter: {
```
zacwest: you'll need a [weak self] here too

Sources/App/Settings/SettingsDetailViewController.swift
```diff
@@ -454,7 +453,26 @@ class SettingsDetailViewController: HAFormViewController, TypedRowControllerType
...
-                    Self.getSceneRows($0)
+                    var baseRows: [BaseRow] = []
+                    if $0 == scenes.first {
```
zacwest: i think it would be better to put a new section at the top for this, or a new RealmSection watching it. this is sort of side-effect-y -- this 'value' needs to update live with contents other than the first one coming in/out, too, right?

zacwest (changes requested):

Sources/App/Settings/SettingsDetailViewController.swift
bgoncal: Not needed anymore after my last commit.

Sources/App/Settings/SettingsDetailViewController.swift
bgoncal: Good one, I forgot to reflect the state on manually changing the scene toggles. It's done now in my latest commit.

bgoncal: Simulator Screen Recording - iPhone 15 Pro - 2023-11-06 at 09 38 51

Sources/App/Settings/SettingsDetailViewController.swift
```diff
@@ -444,6 +445,28 @@ class SettingsDetailViewController: HAFormViewController, TypedRowControllerType
 
             let scenes = realm.objects(RLMScene.self).sorted(byKeyPath: RLMScene.positionKeyPath)
 
+            let toggleAllSwitch = SwitchRow()
+            toggleAllSwitch.title = L10n.SettingsDetails.Actions.Scenes.selectAll
+
+            scenesUpdateObserver = scenes.observe { _ in
+                // Access UISwitch directly to set state to avoid triggering "on value change"
```
zacwest: what happens if you change the model instead, and that triggers a cell update?

zacwest (approved):
if this path doesn't lead anywhere, this version is fine, but i suspect you can change the Eureka model instead

Sources/App/Settings/SettingsDetailViewController.swift
bgoncal: Any changes on scenes model should result in this block of code being executed and updating the "select all" toggle state, is this what you mean or did I misunderstand?

zacwest: Don't put this in one of the model update blocks. Create the model, set up this observer to edit the model, and set the callbacks on the model for when the user toggles it.

bgoncal: I don't think "Select all" should be part of the model, it's just a visual representation that reacts according to the group of scenes state.

bgoncal: I updated the PR based on our talk in discord, also I moved the cell a little bit:
Simulator Screenshot - iPhone 15 Pro - 2023-11-07 at 16 49 07

PR #2451, opened 2023-11-13
Include vocabulary plist for INShareFocusStatusIntent

bgoncal:
Comply to: https://developer.apple.com/documentation/sirikit/registering_custom_vocabulary_with_sirikit/global_vocabulary_reference/intent_phrases

Pending translations to other languages.
Even though INShareFocusStatusIntent is an intent that apparently it's only used for iOS to tell HA when focus is active, we still get warning from Apple about having a siri vocabulary for it, perhaps we can keep it simple like what is in this PR just to address the warning.

### DIFF
--- a/Sources/Extensions/Intents/Resources/Base.lproj/AppIntentVocabulary.plist
+++ b/Sources/Extensions/Intents/Resources/Base.lproj/AppIntentVocabulary.plist
@@ -9,7 +9,7 @@
 			<string>INShareFocusStatusIntent</string>
 			<key>IntentExamples</key>
 			<array>
-				<string>Share focus in home assistant</string>
+				<string>Share focus in Home Assistant</string>
 			</array>
 		</dict>
 	</array>
--- a/Sources/Extensions/Intents/Resources/Base.lproj/AppIntentVocabulary.plist
+++ b/Sources/Extensions/Intents/Resources/Base.lproj/AppIntentVocabulary.plist
@@ -1,0 +1,17 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+<plist version="1.0">
+<dict>
+	<key>IntentPhrases</key>
+	<array>
+		<dict>
+			<key>IntentName</key>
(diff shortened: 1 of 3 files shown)


Sources/Extensions/Intents/Resources/Info.plist
```diff
@@ -47,7 +47,6 @@
 				<string>WidgetActionsIntent</string>
 				<string>WidgetOpenPageIntent</string>
 				<string>INShareFocusStatusIntent</string>
-				<string>AssistIntent</string>
```
bgoncal, line 50: This was duplicated

zacwest (changes requested):
i'm guessing this is like InfoPlist.strings where we need to specify a separate .strings version for translation. can you see if we need to do that? chances are Lokalize won't like getting a .plist like this. either way the scripts to push/pull from Lokalize need to be updated too.

bgoncal:
> i'm guessing this is like InfoPlist.strings where we need to specify a separate .strings version for translation. can you see if we need to do that? chances are Lokalize won't like getting a .plist like this. either way the scripts to push/pull from Lokalize need to be updated too.

I was thinking... should we localise? This string won't be displayed for the user anyways right? Or is this intent visible somewhere?

zacwest:
Does it show up in permission screens or Settings.app at all?

bgoncal:
> Does it show up in permission screens or Settings.app at all?

No, for that we use `NSFocusStatusUsageDescription`

Sources/Extensions/Intents/Resources/Base.lproj/AppIntentVocabulary.plist
```diff
@@ -0,0 +1,17 @@
...
+<plist version="1.0">
+<dict>
+	<key>IntentPhrases</key>
+	<array>
+		<dict>
+			<key>IntentName</key>
+			<string>INShareFocusStatusIntent</string>
+			<key>IntentExamples</key>
+			<array>
+				<string>Share focus in home assistant</string>
```
zacwest: let's do `Home Assistant` even if we're not translating

zacwest (approved):

Sources/Extensions/Intents/Resources/Base.lproj/AppIntentVocabulary.plist
bgoncal: Done

PR #2456, opened 2023-11-27
Thread credentials sharing

bgoncal:
## Summary
Provide a screen to send your thread credentials saved in keychain.

## Screenshots
Screenshot 2023-11-27 at 13 10 08


bgoncal:
Pending:
Properly provide the entitlement for thread management
Remove the button from the thread screen and use a button provided by the web UI

bgoncal:
@zacwest the whole logic for the "ActionButton" may be removed, I had a talk with the web UI team and they will provide a button from the web ui itself

bgoncal:
> @zacwest the whole logic for the "ActionButton" may be removed, I had a talk with the web UI team and they will provide a button from the web ui itself

Confirmed and removed.

Sources/App/Resources/Assets.swift
```diff
@@ -21,10 +21,16 @@ internal typealias AssetImageTypeAlias = ImageAsset.Image
...
+  internal enum Assets {
+    internal static let launchScreenBackground = ColorAsset(name: "launchScreen-background")
+    internal static let launchScreenLogo = ImageAsset(name: "launchScreen-logo")
+    internal static let logo = ImageAsset(name: "Logo")
+    internal static let statusItemIcon = ImageAsset(name: "statusItemIcon")
+  }
+  internal enum Colors {
+    internal static let haPrimary = ColorAsset(name: "haPrimary")
+    internal static let onSurface = ColorAsset(name: "onSurface")
+  }
```
zacwest: can you move these into the 'Shared' target? otherwise, accessing these from e.g. an app extension will not work

Sources/App/Settings/DebugSettingsViewController.swift
```diff
@@ -374,13 +375,29 @@ class DebugSettingsViewController: HAFormViewController {
...
         }
 
+        if #available(iOS 16.4, *) {
+            section <<< ButtonRow {
+                $0.title = L10n.Settings.Developer.MockThreadCredentialsSharing.title
+            }.onCellSelection { [weak self] _, _ in
+                let viewController = UIHostingController(
+                    rootView: ThreadCredentialsSharingView(
+                        viewModel: .init(
+                            server: ServerFixture.standard,
```
zacwest: i think it's fine to use Current.servers.first (or whatnot) for this - no need to include fixtures in non-test-code

Sources/App/Resources/Info.plist
zacwest, line 1: we can revert the changes to this file now right? i don't see any actual changes

Sources/App/Resources/en.lproj/Localizable.strings
```diff
@@ -769,3 +770,9 @@ Home Assistant is free and open source home automation software with a focus on
 "widgets.open_page.not_configured" = "No Pages Available";
 "widgets.open_page.title" = "Open Page";
 "yes_label" = "Yes";
+"thread.credentials.network_name_title" = "Network Name";
+"thread.credentials.network_key_title" = "Network Key";
+"thread.credentials.border_agent_id_title" = "Border Agent ID";
+"thread.credentials.share_credentials_button_title" = "Share credential with Home Assistant";
+"thread.credentials.screen_title" = "Thread Credentials";
+"thread.credentials.error_title" = "Error";
```
zacwest: this one seems unused

Sources/App/Settings/DebugSettingsViewController.swift
```diff
@@ -374,13 +375,29 @@ class DebugSettingsViewController: HAFormViewController {
...
+                            threadClient: MockThreadClientService()
```
zacwest: maybe let's call it 'Simulator…' rather than 'Mock…' since it is not really mocking anything

Sources/App/DesignSystem/Components/HAButton.swift
```diff
@@ -0,0 +1,30 @@
+import SwiftUI
+
+@available(iOS 13, *)
+public struct TextButton: ButtonStyle {
+    public func makeBody(configuration: Configuration) -> some View {
```
zacwest, line 7: can you give this button a highlighted state?

Sources/App/ThreadCredentialsSharing/ThreadClientService.swift
```diff
@@ -0,0 +1,61 @@
...
+                channel: credential.channel,
+                panID: credential.panID?.hexadecimal ?? placeholder,
+                creationDate: credential.creationDate,
+                lastModificationDate: credential.lastModificationDate
+            )
+        }
+    }
+}
+
+extension Data {
```
zacwest: private? or move to Shared?

Sources/App/ThreadCredentialsSharing/ThreadCredentialsSharingView.swift
```diff
@@ -0,0 +1,134 @@
...
+            )
+            makeCardPropertyView(
+                title: L10n.Thread.Credentials.networkKeyTitle,
+                value: credential.networkKey
+            )
+            Button {
+                viewModel.shareCredentialWithHomeAssistant(credential: credential)
+            } label: {
+                Text(L10n.Thread.Credentials.shareCredentialsButtonTitle)
+            }
```
zacwest, line 118: can this button get a loading indicator while the network request is executing, or make it disabled?

can we get a success HUD of some sort, or alert?

Sources/App/ThreadCredentialsSharing/ThreadCredentialsSharingView.swift
```diff
@@ -0,0 +1,134 @@
...
+        }
+    }
+
+    private func makeCardPropertyView(title: String, value: String) -> some View {
+        VStack(alignment: .leading, spacing: .zero) {
+            Group {
+                Text(title)
+                    .font(.footnote)
+                Text(value)
+                    .textSelection(.enabled)
```
zacwest, line 129: thank you for doing this :)

Sources/Shared/Environment/Environment.swift
```diff
@@ -365,6 +365,19 @@ public class AppEnvironment {
 
     public var matter = MatterWrapper()
 
+    public var threadCredentialsSharingEnabled: Bool {
```
zacwest: i know it's kind of mixing different layers, but…how do you feel about putting this in MatterWrapper?

Sources/App/WebView/WebViewController.swift
```diff
@@ -1096,6 +1101,18 @@ extension WebViewController: WKScriptMessageHandler {
             }
         }
     }
+
+    private func threadCredentialsRequested() {
+        if #available(iOS 16.4, *) {
+            guard let server = Current.settingsStore.menuItemTemplate?.server else { return }
```
zacwest: `self.server` -- every WebViewController is bound to exactly 1 server, and it's non-optional

zacwest (changes requested):
the credential sharing screen seems to be wrapped in a split view in landscape, making the source list hidden by default.

simulator_screenshot_F1A8C04B-9EC2-44E0-9F5D-4D31F69A62BA
simulator_screenshot_77FA6E14-4AD1-41FA-B167-E3F21E67716F

Sources/App/ThreadCredentialsSharing/ThreadCredentialsSharingView.swift
bgoncal: There is a success alert already, I won't change this for now because today I have a meeting with the designer so there is a chance more in terms of UI will be changed, but it can be an iteration after merging this PR

Sources/Shared/Environment/Environment.swift
bgoncal: Good point, I think in the future we need to separate them, since they are different concerns, but for now I will move it inside MatterWrapper while we don't have more from thread

swiftgen.yml
```diff
@@ -27,7 +27,9 @@ xcassets:
     - Sources/App/Resources/Colors.xcassets
   outputs:
     templateName: swift5
-    output: Sources/App/Resources/Assets.swift
+    output: Sources/Shared/Assets/Assets.swift
```
zacwest, line 30: you'll need to move the .xcassets into Shared as well, or else anything-but-the-main-app will not find the given asset name

zacwest (changes requested):
just need to relocate the xcassets otherwise lgtm

swiftgen.yml
bgoncal: I updated to have one shared asset catalog in shared target and one asset catalog on the main target, each one for what makes more sense. I also replaced the app icon by the single-asset app icon as we talked.

zacwest (changes requested):
Assets.xcassets is now in both the App target and Shared target, i think you need to break it apart:

1. the app icon xcassets (which i think _has_ to be in the main app) in the main app
2. the shared assets (color or images) in shared/ only

having them in both means we're duplicating it unnecessarily

bgoncal:
> Assets.xcassets is now in both the App target and Shared target, i think you need to break it apart:
> 
> 1. the app icon xcassets (which i think _has_ to be in the main app) in the main app
> 2. the shared assets (color or images) in shared/ only
> 
> having them in both means we're duplicating it unnecessarily

Ok so I updated and now we have "Assets" in main target and "SharedAssets" in shared, Colors I kept just one inside shared for now.

PR #2466, opened 2023-12-03
Avoid CI to have duplicate runs if changes are pushed in a row

bgoncal:

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -16,6 +16,9 @@ env:
 
 jobs:
   lint:
+    concurrency:
+      group: ci-${{ github.ref }}
+      cancel-in-progress: true
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -16,6 +16,9 @@ env:
 
 jobs:
   lint:
+    concurrency:
+      group: ci-lint-${{ github.ref }}
+      cancel-in-progress: true
```
zacwest: would it be better to use a `ci.yml`-level concurrency value? (e.g. not in the individual jobs)

zacwest (approved):

.github/workflows/ci.yml
bgoncal: I tried that but it was complaining that I needed to put under a job.
Before merging I'll try again, that would be better indeed.

zacwest: oh weird, [the docs](https://docs.github.com/en/actions/using-jobs/using-concurrency#example-using-concurrency-and-the-default-behavior) really suggest you can do it globally in the file

bgoncal: I found my mistake, I was placing under jobs, thats why the warning, now it should be good

PR #2469, opened 2023-12-05
Drop iOS 12, 13 and 14 support

bgoncal:
## Summary

Drop support for iOS 12 and 13.


HomeAssistant.xcodeproj/project.pbxproj
```diff
@@ -6522,6 +6522,7 @@
 				CODE_SIGN_ENTITLEMENTS = "Configuration/Entitlements/Extension-ios.entitlements";
 				"CODE_SIGN_ENTITLEMENTS[sdk=macosx*]" = "Configuration/Entitlements/Extension-catalyst.entitlements";
 				INFOPLIST_FILE = Sources/Extensions/Share/Resources/Info.plist;
+				IPHONEOS_DEPLOYMENT_TARGET = 15.0;
```
zacwest: this should be set in the xcconfig, rather than the xcodeproj

(when in doubt, do it in the xcconfig - makes reviewing/history a lot easier, and reduces a lot of churn in the proj file)

HomeAssistant.xcodeproj/project.pbxproj
```diff
@@ -6632,7 +6635,7 @@
 				CODE_SIGN_ENTITLEMENTS = "Configuration/Entitlements/Extension-ios.entitlements";
 				"CODE_SIGN_ENTITLEMENTS[sdk=macosx*]" = "Configuration/Entitlements/Extension-catalyst.entitlements";
 				INFOPLIST_FILE = Sources/Extensions/Widgets/Resources/Info.plist;
-				IPHONEOS_DEPLOYMENT_TARGET = 14.0;
+				IPHONEOS_DEPLOYMENT_TARGET = 15.0;
```
zacwest: you can probably remove the target override now for iOS 14 since we're dropping it, it was mostly here because it was newer

Sources/App/AppDelegate.swift
```diff
@@ -208,55 +191,19 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
...
-            }
-
-            if NSClassFromString("XCTest") != nil {
-                return false
-            }
-
-            Current.Log.info("allowing state to be restored")
-            return true
-        }
+        false
```
zacwest: i think you can delete the whole method now

Sources/App/AppDelegate.swift
```diff
@@ -208,55 +191,19 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
...
-            if Current.settingsStore.restoreLastURL == false {
-                // if we let it capture state -- even if we don't use the url -- it will take a screenshot
-                Current.Log.info("disallowing state to be saved due to setting")
-                return false
-            }
-
-            Current.Log.info("allowing state to be saved")
-            return true
-        }
+        false
```
zacwest: same same

Sources/App/AppDelegate.swift
```diff
@@ -208,55 +191,19 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
...
         coder: NSCoder
     ) -> UIViewController? {
-        if #available(iOS 13, *) {
-            return nil
-        } else {
-            return sceneManager.compatibility.windowController?.viewController(
-                withRestorationIdentifierPath: identifierComponents
-            )
-        }
+        nil
```
zacwest: probably delete it all

Sources/App/Settings/ActionConfigurator.swift
```diff
@@ -115,12 +113,9 @@ class ActionConfigurator: HAFormViewController, TypedRowControllerType {
             }
         }
 
-        if #available(iOS 13.0, *) {
-            // after text if uneditable
```
zacwest, line 119: preserve the comment?

Sources/App/WebView/WebViewController.swift
```diff
@@ -30,11 +30,7 @@ final class WebViewController: UIViewController, WKNavigationDelegate, WKUIDeleg
         withRestorationIdentifierPath identifierComponents: [String],
         coder: NSCoder
     ) -> UIViewController? {
-        if #available(iOS 13, *) {
-            return nil
-        } else {
-            return WebViewController(restoring: .coder(coder))
-        }
+        nil
```
zacwest: i think you can delete this whole method

zacwest (changes requested):

HomeAssistant.xcodeproj/project.pbxproj
bgoncal: I updated the xcconfig but I don't think it sets the deployment version for all extensions right? I had to manually change some

zacwest: Yeah, some were overridden to be newer, but all less than iOS 15. We should be able to delete all xcodeproj-level deployment targets.

HomeAssistant.xcodeproj/project.pbxproj
zacwest, line 1: all the green stuff in here can probably be deleted

zacwest (approved):

Cyberbeni:
The patch notes say "more information is available when you open the App" but I didn't see anything special after opening the app. There is also no proper description for this pull request. There are no devices that have iOS 13/14 as their last OS, so I get that they can be dropped but I know multiple people who still use devices that only have iOS 12.

PR #2488, opened 2023-12-21
Move alternate icons to asset catalog

bgoncal:
## Summary

Alternate icons were in a project folder, I moved them to assets catalog and removed uneeded references from info.plist


Sources/App/Settings/SettingsDetailViewController.swift
```diff
@@ -867,8 +878,8 @@ enum AppIcon: String, CaseIterable {
     case Purple = "purple"
     case Red = "red"
     case White = "white"
-    case BiPride = "bi_pride"
```
zacwest: what happens to users who had the old name set as their icon?

zacwest (changes requested):
nice that emerge caught this. i think you need to use a different asset catalog as we do not want to actually include the originals in the app.

Sources/App/Settings/SettingsDetailViewController.swift
bgoncal: Good catch, I have reverted the naming

bgoncal:
> nice that emerge caught this. i think you need to use a different asset catalog as we do not want to actually include the originals in the app.

I don't see why not; previously, we also included the original assets in the app, right? The only difference is that now they were moved to an asset catalog instead of just images on a folder and with a single size, which is bigger than what we previously had (no clue why but the previous assets were blurry, it could have been an issue when used on iPad Pro 12,9")

zacwest:
We did not include the 1024x1024 icons, just the resized ones; hence why the app size went up by 8 MB.

bgoncal:
> We did not include the 1024x1024 icons, just the resized ones; hence why the app size went up by 8 MB.

But for the single size asset we need a 1024x1024 icon and in theory it's handle by Apple the resizing, how a different asset catalog would improve that? Or you mean we should not use the single size asset and go back to the old way specifying each icon size?

zacwest:
When you tell it the alternate asset catalog to use, it only includes the resized icons and not the full ones. By keeping it in the main asset catalog you are also causing it to include the 1024 which isn't used by the system at runtime.

HomeAssistant.xcodeproj/project.pbxproj
```diff
@@ -6923,6 +6928,7 @@
 			isa = XCBuildConfiguration;
 			baseConfigurationReference = 675CE4281FE5F1920B13D553 /* Pods-iOS-App.debug.xcconfig */;
 			buildSettings = {
+				ASSETCATALOG_COMPILER_ALTERNATE_APPICON_NAMES = "trans beta dev black caribbean-green blue cornflower-blue electric-violet crimson green classic old-beta old-dev old-release fire-orange orange purple pink white red bi_pride release POC_pride non-binary rainbow";
```
bgoncal: @zacwest I couldn't find a way to specify the asset catalog itself, if you have done that before let me know the best way

zacwest: i can't seem to find it either. i think you are right that `ASSETCATALOG_COMPILER_INCLUDE_ALL_APPICON_ASSETS` is a better approach. how about - let's push up a testflight build and see if the app store app sizing says it's different or not. maybe the AS will strip the 1024x1048 file?

bgoncal: Sounds good, I reverted my change, let's see how it goes

Issue #1040, opened 2020-09-18
App does not remember the window size


domoritz:
**Device model, version and app version**

Software Version: macOS 11
App version: 2020.7 (1)

**Describe the bug**

The window does not remember the window size across closing/opening the app. 

**To Reproduce**

* Resize window
* Close app
* Open app again

**Expected behavior**

The window should remember its previous size.

zacwest:
Do you have System Preferences > General > "Close windows when quitting an app" enabled?

domoritz:
Yes, I do.

SeanPM5:
Had the same behavior here since a few builds, I thought it was just me. It was always opening in the top left of the screen in like a 600px window.

That "Close windows when quitting an app" option was on for me (never changed it), after unchecking it it's now remembering the window size.

Also on macOS 11 / Big Sur PB.

zacwest:
Multi-window support means it's now persisting information about the windows when quitting, and if there are none, it doesn't have anything. I need to investigate if the Catalyst side of things will allow us to manually do so.

carsten-h:
The option "System Preferences > General > "Close windows when quitting an app"" is not longer there and the positions are not saved!
This information is for Home Assistant Companion 2021.1 (1).

zacwest:
<img width="780" alt="image" src="https://user-images.githubusercontent.com/74188/103182395-d22a1b80-485f-11eb-93a2-eaee76952806.png">

carsten-h:
Oh, you are right, I totally read over "System"!

wojciechpolak:
I unchecked the system option "Close windows when quitting an app" and now when I quit the HA desktop app v2021.1.1 (9) it remembers its custom window size & position. However, when I click the top-left red circle close button it closes the app without quitting it AND THEN when I click the app again from the dock, its previous window size & position is gone and reset back to default.

skrenes:
I have my app setup with the "Menu Bar" option. Whenever I close the window and then tap the menu bar icon, it goes back to the default small window size centered on my first display.
<img src="https://user-images.githubusercontent.com/5075909/113428730-8cb44780-9394-11eb-8f78-9dbc4096af83.png" width="50%">

matthew-larner:
"System Preferences > General > "Close windows when quitting an app" isn't the best way to handle this issue. It should remember the window size and position when closing the window irrespective of this setting.

zacwest:
[...]

PR #2514, opened 2024-01-08
Restore mac window size when reopen the app

bgoncal:
## Summary

Restore mac catalyst app latest window size when reopening the app


bgoncal:
As reported on https://github.com/home-assistant/iOS/issues/1040

Sources/App/Scenes/WindowScenesManager.swift
```diff
@@ -0,0 +1,75 @@
...
+
+        let numberOfConnectedScenes = UIApplication.shared.connectedScenes.count
+        let adjustedSystemFrame = adjustedSystemFrame(
+            preferredSystemFrame,
+            for: screenSize,
+            numberOfConnectedScenes: numberOfConnectedScenes
+        )
+
+        #if targetEnvironment(macCatalyst)
+        scene.requestGeometryUpdate(.Mac(systemFrame: adjustedSystemFrame)) { error in
```
zacwest, line 71: what happens if the system is restoring a window with a customized size? we shouldn't reset all of them, just newly created (not restored) scenes

zacwest (changes requested):

Sources/App/Scenes/WindowScenesManager.swift
bgoncal: When would that happen? How can I reproduce?
Currently this rules is applied for the webview scene, the size it left when you closed it is the size it will reopen

zacwest: Open a few windows, resize them, quit the app and reopen the app. State restoration will open the windows again unless you have the macOS-level "close files when quitting an app" setting on.

bgoncal: Looks good right?

https://github.com/home-assistant/iOS/assets/5808343/b47ac0ca-d7d7-469a-bcd6-ab15735595c2

zacwest: i feel like the logic here is dependent on timing (that connected scenes exist in multiples before this method is invoked) that isn't guaranteed. we can tell if a scene is restoring state (if it had a user activity, basically) - can we use that to decide not to do this?

bgoncal: Yup, I've added a check, let me know if it is enough

Sources/App/Scenes/WindowScenesManager.swift
```diff
@@ -0,0 +1,77 @@
...
+    private func sceneFrameIsValid(_ sceneFrame: CGRect, screenSize: CGSize) -> Bool {
+        sceneFrame.height <= screenSize.height && sceneFrame.width <= screenSize.width
+    }
+
+    private func adjustedSystemFrame(
+        _ systemFrame: CGRect,
+        for screenSize: CGSize,
+        numberOfConnectedScenes: Int
+    ) -> CGRect {
+        guard numberOfConnectedScenes > 1 else { return systemFrame }
```
zacwest, line 34: what does this do? can you add a comment?

Sources/App/Scenes/WindowScenesManager.swift
```diff
@@ -0,0 +1,77 @@
+import Foundation
+import Shared
+import UIKit
+
+final class WindowScenesManager {
```
zacwest, line 5: i feel like this class may be worth writing some tests around

bgoncal: Updated

Sources/App/Scenes/WindowScenesManager.swift
zacwest: i'm still unsure what the connected scene count is doing here. why not restore the frame if there's 1 connected scene?

bgoncal: So basically, if you have just one scene, ir will recover the latest window size, if you have more than one scene and both were positioned in the same coordinates, it will make them a little bit offset (cascade effect) so one is not hidden.
I got this idea from one article and I thought it would be a nice touch, what do you think?

zacwest: ahhh i understand now. yeah that is a good effect!

ios-52:
Hello
I have restore successfully window size but problem is when i reopen app than view design cut in top side.can any one help this.

https://github.com/user-attachments/assets/5656175b-e73a-4f02-ad8d-ee57fbf8a670

PR #2517, opened 2024-01-09
Add CarPlay Actions and general improvements

bgoncal:
## Summary

- Replace CarPlay navigation by tabs
- Add iOS Actions

## Screenshots

Screenshot 2024-01-09 at 15 12 21


Sources/App/Scenes/CarPlaySceneDelegate.swift
```diff
@@ -13,114 +13,56 @@ public protocol EntitiesStateSubscription {
...
-            if self.serverId == nil {
-                /// No server is selected
-                guard let server = self.getServer() else {
-                    Current.Log.info("No server connected")
-                    return
-                }
-                self.setServer(server: server)
-            }
-        }
+    private let realm = Current.realm()
```
zacwest: this is designed to be called each time (it is thread specific, etc.) and Current doesn't really want you to cache its values

HomeAssistant.xcodeproj/project.pbxproj
```diff
@@ -502,6 +502,7 @@
 		420B100C2B1D204400D383D8 /* Assets.xcassets in Resources */ = {isa = PBXBuildFile; fileRef = 420B100B2B1D204400D383D8 /* Assets.xcassets */; };
 		424A7F462B188946008C8DF3 /* WidgetBackground.swift in Sources */ = {isa = PBXBuildFile; fileRef = 424A7F452B188946008C8DF3 /* WidgetBackground.swift */; };
 		424A7F482B188BF3008C8DF3 /* WidgetContentMargin.swift in Sources */ = {isa = PBXBuildFile; fileRef = 424A7F472B188BF3008C8DF3 /* WidgetContentMargin.swift */; };
+		424DD05A2B3509170057E456 /* CPActionsTemplate.swift in Sources */ = {isa = PBXBuildFile; fileRef = 424DD0592B3509170057E456 /* CPActionsTemplate.swift */; };
```
zacwest: let's not use CP as a prefix for our own things - apple generally advises against it but it can also make it hard to tell what is a system thing or not (or may clash later, in objc land). maybe HCP or something if we wanna prefix?

Sources/App/Scenes/CarPlaySceneDelegate.swift
```diff
@@ -13,114 +13,56 @@ public protocol EntitiesStateSubscription {
...
-            server: server
-        )
-
-        guard let domainsListTemplate else { return }
-
-        domainsListTemplate.interfaceController = interfaceController
+    private func setupActionsTemplate() {
+        let actions = realm.objects(Action.self)
+            .sorted(byKeyPath: "Position")
+            .filter("Scene == nil")
```
zacwest: any reason to exclude scene actions?

Sources/Vehicle/Templates/CPServerListTemplate.swift
```diff
@@ -0,0 +1,99 @@
...
+        template.tabTitle = L10n.Carplay.Labels.servers
+        template.tabImage = MaterialDesignIcons.cogIcon.image(ofSize: .init(width: 64, height: 64), color: nil)
+    }
+
+    private func serverListTemplate() -> CPTemplate {
+        var serverList: [CPListItem] = []
+        for server in Current.servers.all {
+            let serverItem = CPListItem(
+                text: server.info.name,
+                detailText: "\(server.info.connection.activeURLType.description) - \(server.info.connection.activeURL().absoluteString)"
```
zacwest: do we need this info in the subtitle? we tend not to show urls - just things like account name. maybe copy the title logic from settings?

Sources/Vehicle/Templates/CPServerListTemplate.swift
```diff
@@ -0,0 +1,99 @@
...
+        return CPListTemplate(title: L10n.Carplay.Labels.servers, sections: [section])
+    }
+
+    private func setServer(server: Server) {
+        serverId = server.identifier
+        prefs.set(server.identifier.rawValue, forKey: CPServersListTemplate.carPlayPreferredServerKey)
+    }
+
+    /// Get server for ID or first server available
+    private func getServer(id: Identifier<Server>? = nil) -> Server? {
```
zacwest: i think this is a method on the server manager already, to make the logic for "which server?" cleaner in things like notification actions

Sources/Vehicle/Templates/CPActionsTemplate.swift
```diff
@@ -0,0 +1,108 @@
...
+    }
+
+    func templateWillAppear(template: CPTemplate) {
+        if template == self.template {
+            update()
+        }
+    }
+
+    func update() {
+        template = list(for: actions)
```
zacwest: is there a reason to create new templates each time rather than updating the old one?

Sources/Vehicle/Templates/CPActionsTemplate.swift
```diff
@@ -0,0 +1,108 @@
...
+        template.tabTitle = L10n.Carplay.Navigation.Tab.actions
+        template.tabImage = MaterialDesignIcons.lightningBoltIcon.image(
+            ofSize: .init(width: 64, height: 64),
+            color: nil
+        )
+        template.tabSystemItem = .more
+    }
+
+    func list(for actions: Results<Action>) -> CPTemplate {
+        actionsToken = actions.observe { [weak self] _ in
```
zacwest: maybe reset the previous one just in case the call sites change to make this happen unexpectedly

Sources/Vehicle/Templates/CPActionsTemplate.swift
zacwest: is there a reason to create a new template each time? i think it might make these objects a little easier to follow if you update in all cases rather than reset in some

Sources/App/Scenes/CarPlaySceneDelegate.swift
bgoncal: We are having some talks internally and the idea is to start small, not bring too much information to CarPlay, but thats a good point, I'll follow up

HomeAssistant.xcodeproj/project.pbxproj
bgoncal: I replaced by "CarPlay..."

Sources/Vehicle/Templates/CPServerListTemplate.swift
bgoncal: Agree, tmi, I will leave the server name only, since it should be enough for the user to choose between servers

Sources/Vehicle/Templates/CPServerListTemplate.swift
bgoncal: Updated

Sources/Vehicle/Templates/CPActionsTemplate.swift
bgoncal: Not really, updated.

bgoncal: Not really, updated.

Sources/App/Scenes/CarPlaySceneDelegate.swift
bgoncal: We are keeping only actions for now, until we get some feedback

Sources/App/Scenes/CarPlaySceneDelegate.swift
```diff
@@ -173,49 +98,20 @@ extension CarPlaySceneDelegate: CPTemplateApplicationSceneDelegate {
...
-            return
-        }
-        setServer(server: server)
     }
 }
 
 @available(iOS 16.0, *)
 extension CarPlaySceneDelegate: CPInterfaceControllerDelegate {
     func templateWillDisappear(_ aTemplate: CPTemplate, animated: Bool) {
         domainsListTemplate?.templateWillDisappear(template: aTemplate)
```
zacwest: feels like a "2 or more, use a for" situation. maybe just use an array to do this and the opposite transition?

Sources/Shared/API/ServerManager.swift
```diff
@@ -11,6 +11,7 @@ public protocol ServerObserver: AnyObject {
 public protocol ServerManager {
     var all: [Server] { get }
     func server(for identifier: Identifier<Server>) -> Server?
+    func server(for identifier: Identifier<Server>, fallback: Bool) -> Server?
```
zacwest: since this one is going to behave slightly different than the other fallbacks, can we rename it so it implies more of a 'guaranteed'-ish return?

Sources/Vehicle/Templates/CarPlayTemplateProvider.swift
```diff
@@ -0,0 +1,10 @@
+import CarPlay
+import Foundation
+
+protocol CarPlayTemplateProvider {
+    var template: CPTemplate { get set }
```
zacwest: might not be a bad idea to make this (if you can) `some CPTemplate` (and `any CarPlayTemplateProvider`) so you can specialize each usage of it, so they don't have to cast internally

Sources/Vehicle/Templates/CarPlayActionsTemplate.swift
```diff
@@ -0,0 +1,107 @@
...
+    func updateList(for actions: Results<Action>) {
+        actionsToken?.invalidate()
+        actionsToken = actions.observe { [weak self] _ in
+            self?.updateActions(actions: actions)
+        }
+
+        if let listTemplate = template as? CPListTemplate {
+            listTemplate.updateSections([section(actions: actions)])
+        } else {
+            template = CPListTemplate(
```
zacwest: since the template was taken in the updateTemplates method in the scene delegate, how does modifying it here trigger a re-set in the tab bar?

Sources/Vehicle/Templates/CarPlayActionsTemplate.swift
```diff
@@ -0,0 +1,107 @@
...
+import Foundation
+import RealmSwift
+import Shared
+
+@available(iOS 15.0, *)
+final class CarPlayActionsTemplate: CarPlayTemplateProvider {
+    private var actionsToken: NotificationToken?
+    private let actions: Results<Action>
+
+    private var noActionsView: CPInformationTemplate = {
```
zacwest: unused

Sources/App/Scenes/CarPlaySceneDelegate.swift
```diff
@@ -13,114 +13,55 @@ public protocol EntitiesStateSubscription {
...
-        updateServerListButton()
-    }
-
-    private func updateServerListButton() {
-        domainsListTemplate?.setServerListButton(show: Current.servers.all.count > 1)
+    private var domainsListTemplate: CarPlayTemplateProvider?
+    private var serversListTemplate: CarPlayTemplateProvider?
+    private var actionsListTemplate: CarPlayTemplateProvider?
+
+    @objc private func updateTemplates() {
```
zacwest: feels like a lot of this doesn't necessarily need to reset each time. we're not really treating them as immutable, so there's a quasi-immutable thing going on here. maybe it would be better to strongly declare the template providers on init -- when changing server, tell them about which one is selected (if they care); but otherwise let them update responsively.

zacwest (changes requested):
mostly for this i think the scene delegate needs a little bit of abstraction (making the providers an array, perhaps). otherwise i think this is definitely in the right direction.

Sources/App/Scenes/CarPlaySceneDelegate.swift
bgoncal: Agree, I've done some clean up

Sources/Vehicle/Templates/CarPlayActionsTemplate.swift
bgoncal: Updated it, it shouldn't matter anymore

Sources/Vehicle/Templates/CarPlayTemplateProvider.swift
bgoncal: Good one!

Sources/App/Scenes/CarPlaySceneDelegate.swift
```diff
@@ -13,113 +13,42 @@ public protocol EntitiesStateSubscription {
...
-    }
-
-    private func updateServerListButton() {
-        domainsListTemplate?.setServerListButton(show: Current.servers.all.count > 1)
+    private var domainsListTemplate: any CarPlayTemplateProvider
+    private var serversListTemplate: any CarPlayTemplateProvider
+    private var actionsListTemplate: any CarPlayTemplateProvider
+    private var areasZonesListTemplate: any CarPlayTemplateProvider
+
+    private lazy var allTemplates: [any CarPlayTemplateProvider] = {
```
zacwest: probably fine to not make this stored (`{ [ a, b, c ] }` not `= { […] }`)

Sources/Vehicle/Templates/CarPlayPaginatedListTemplate.swift
```diff
@@ -0,0 +1,98 @@
...
+
+    private func getPageButtons(endIndex: Int, currentPage: Int, totalCount: Int) -> [CPBarButton] {
+        var barButtons: [CPBarButton] = []
+
+        guard let forwardImage = UIImage(systemName: "arrow.forward"),
+              let backwardImage = UIImage(systemName: "arrow.backward") else { return [] }
+
+        if endIndex < totalCount {
+            barButtons.append(CPBarButton(
+                image: forwardImage,
```
zacwest, line 61: i don't know why i am surprised that there's no ability to set a title too. i guess they assume a car operator is sighted, but what about a passenger using it? oh well

Sources/Vehicle/Templates/CarPlayPaginatedListTemplate.swift
```diff
@@ -0,0 +1,98 @@
...
+                handler: { _ in
+                    self.changePage(to: .next)
+                }
+            ))
+        } else {
+            barButtons
+                .append(CPBarButton(
```
zacwest, line 68: does this look better than not including it?

Sources/Vehicle/Templates/CarPlayPaginatedListTemplate.swift
```diff
@@ -0,0 +1,98 @@
...
+        var barButtons: [CPBarButton] = []
+
+        guard let forwardImage = UIImage(systemName: "arrow.forward"),
+              let backwardImage = UIImage(systemName: "arrow.backward") else { return [] }
+
+        if endIndex < totalCount {
+            barButtons.append(CPBarButton(
+                image: forwardImage,
+                handler: { _ in
+                    self.changePage(to: .next)
```
zacwest: in modern swift there's not a lot of reasons to do `self.` in a closure, choose one of:

```swift
{ [self] in
  changePage(to: next)
}
```

(to strongly retain self) or

```swift
{ [weak self] in
  guard let self else { return }
  changePage(to: next)
}
```

(to weakly retain self)

Sources/Vehicle/Templates/CarPlayPaginatedListTemplate.swift
```diff
@@ -0,0 +1,98 @@
...
+    private(set) var template: CPListTemplate
+
+    init(title: String, items: [CPListItem]) {
+        self.title = title
+        self.items = items
+        self.currentPage = 0
+        self.template = CPListTemplate(title: title, sections: [])
+    }
+
+    func updateItems(items: [CPListItem], refreshUI: Bool = false) {
```
zacwest, line 31: what is the purpose of 'refreshUI' as an option here?

Sources/Vehicle/Templates/CarPlayEntitiesListTemplate.swift
```diff
@@ -5,65 +5,112 @@ import PromiseKit
...
             }
         }
-
-        return template
     }
 
-    private func updateListItems() {
-        guard let entities = entitiesCachedStates.value, let listTemplate = template as? CPListTemplate else { return }
+    func update() {
+        guard let entities = entitiesCachedStates.value else { return }
```
zacwest: most all of this logic can be moved into a `map` of the `entitesCachedStates` turning this into a `HACache<[HAEntity]>`

Sources/Vehicle/Templates/CarPlayEntitiesListTemplate.swift
```diff
@@ -5,65 +5,112 @@ import PromiseKit
...
 
-        let startIndex = currentPage * itemsPerPage
-        let endIndex = min(startIndex + itemsPerPage, entitiesSorted.count)
-        let entitiesToShow = Array(entitiesSorted[startIndex ..< endIndex])
+            return (e1.attributes.friendlyName ?? e1.entityId) < (e2.attributes.friendlyName ?? e2.entityId)
+        })
+
+        // Prevent unecessary update and UI glitch for non-touch screen CarPlay
+        let entitiesIds = entitiesSorted.map(\.entityId).sorted()
+        guard entitiesIdsCurrentlyInList != entitiesIds else {
```
zacwest: does this ever match when pagination is required?

i think it may be easier to match the entities here in a dictionary to a `CPListItem`, e.g.

```
// at a class level
var entityToListItem: [String: CPListItem]

// in this method
for entity in entities {
  let listItem: CPListItem
  if let existing = entityToListItem[entity.entityId] {
    listItem = existing
  } else {
    listItem = CPListItem()
    entityToListItem[entity.entityId] = listItem
  }
  // do updates / configure

}
```

i think you would need to move the "and change the items displayed" into the paginator to really invalidate/reset the template

Sources/Vehicle/Templates/CarPlayPaginatedListTemplate.swift
```diff
@@ -0,0 +1,98 @@
...
+                .append(CPBarButton(
+                    image: UIImage(size: backwardImage.size, color: UIColor.clear),
+                    handler: nil
+                ))
+        }
+
+        return barButtons
+    }
+
+    private func changePage(to: GridPage) {
```
zacwest, line 92: what happens if you spam the button and this gets fired twice at the end, somehow?

Sources/Vehicle/Templates/CarPlayServerListTemplate.swift
```diff
@@ -0,0 +1,82 @@
...
+    }
+
+    init() {
+        self.template = CPListTemplate(title: "", sections: [])
+        template.tabTitle = L10n.Carplay.Labels.servers
+        template.tabImage = MaterialDesignIcons.cogIcon.carPlayIcon(color: nil)
+    }
+
+    func templateWillDisappear(template: CPTemplate) {
+        Current.servers.remove(observer: self)
```
zacwest: this isn't checking the template, is that an issue?

Sources/App/Scenes/CarPlaySceneDelegate.swift
```diff
@@ -173,49 +86,16 @@ extension CarPlaySceneDelegate: CPTemplateApplicationSceneDelegate {
...
-        }
-        setServer(server: server)
     }
 }
 
 @available(iOS 16.0, *)
 extension CarPlaySceneDelegate: CPInterfaceControllerDelegate {
     func templateWillDisappear(_ aTemplate: CPTemplate, animated: Bool) {
-        domainsListTemplate?.templateWillDisappear(template: aTemplate)
+        allTemplates.forEach { $0.templateWillDisappear(template: aTemplate) }
```
zacwest, line 95: should these check if the template matches the individual one before calling? it would help eliminate the 'if template == self.template' check in (most?) of the implementations

zacwest (changes requested):

Sources/App/Scenes/CarPlaySceneDelegate.swift
bgoncal: I was implementing that when I remember some templates have child templates (like the domain list > entities list and the Areas list > entities list).
Those we don't have the reference in scene delegate so this "event" needs to get to them from their parents

Sources/Vehicle/Templates/CarPlayServerListTemplate.swift
bgoncal: Yup, I'll fix

Sources/Vehicle/Templates/CarPlayPaginatedListTemplate.swift
bgoncal: You can set a title OR an image, I think it's due to space, in this case Icons look better

Sources/Vehicle/Templates/CarPlayPaginatedListTemplate.swift
bgoncal: In my pinion it does, it prevents the icons to move to a different position when you have just one possible action (forward or backward), if this is not included the user could keep hitting the icon that is in the same position because it thinks it's the same actions while it is not:

https://github.com/home-assistant/iOS/assets/5808343/a8b121cf-55ae-4ff1-a377-678d6d39f208

Sources/Vehicle/Templates/CarPlayPaginatedListTemplate.swift
bgoncal: Refresh UI would literally update all items from the list, which causes a glitch that scrolls the user to the top and move the cursor (for non-touchscreen carplays), so this is only used when you are setting the list for the first time, I will add a comment to explain that

Sources/Vehicle/Templates/CarPlayEntitiesListTemplate.swift
bgoncal: That is sort of what is happening already, the the list of entityIds change it means that some entity was added or removed, which means the paginated list needs an update, otherwise if the entityIds list is the same it means the state of an or multiple entities were updated, then those specific items are updated IF they are visible at that moment.

in CPListTemplate there is also not the concept of inserting a new item, so when we have a new entity we have to update the whole section

Sources/Vehicle/Templates/CarPlayPaginatedListTemplate.swift
bgoncal: I couldn't reproduce it neither on simulator or my car, so it would be a very edge case I guess.

bgoncal:
Current state
Screenshot 2024-01-15 at 12 05 39

Sources/Vehicle/Templates/CarPlayEntitiesListTemplate.swift
zacwest: I see the intent but I think pagination is making this "if" never work for cases where pagination happens at all.

bgoncal: This check happens before everything is paginated, it's goal is just to prevent to recreate the list when the only thing that has changed was the state of entities. Let's say a new entity is added to your HA instance, then entitiesIds will have one more item and therefore different from entitiesIdsCurrentlyInList (previously set), which then will lead to the list being recreated.

PR #2541, opened 2024-01-24
Improve iOS Actions, facilitate automation creation

bgoncal:
## Summary

Reorder sections in action configurator and added button to create automation directly from action configurator

## Screenshots

Simulator Screenshot - iPhone 15 Pro - 2024-01-24 at 22 44 38

### DIFF
--- a/Sources/App/Settings/ActionConfigurator.swift
+++ b/Sources/App/Settings/ActionConfigurator.swift
@@ -18,9 +18,10 @@ class ActionConfigurator: HAFormViewController, TypedRowControllerType {
         }
     }
 
-    var newAction: Bool = true
-    var shouldSave: Bool = false
-    var preview = ActionPreview(frame: CGRect(x: 0, y: 0, width: 169, height: 44))
+    private var newAction: Bool = true
+    private(set) var shouldSave: Bool = false
+    private(set) var shouldOpenAutomationEditor: Bool = false
+    private var preview = ActionPreview(frame: CGRect(x: 0, y: 0, width: 169, height: 55))
 
     convenience init(action: Action?) {
         self.init()
@@ -52,6 +53,18 @@ class ActionConfigurator: HAFormViewController, TypedRowControllerType {
             }
         }
 
+        form +++ ViewRow<ActionPreview>("preview") { [weak self] row in
+            row.hidden = Condition.function(["showInWatch"], { _ in
+                !(self?.action.showInWatch ?? true)
+            })
+        }.cellSetup { [weak self] cell, _ in
+            guard let self else { return }
+            cell.backgroundColor = UIColor.clear
+            cell.preservesSuperviewLayoutMargins = false
+            self.updatePreviews()
+            cell.view = self.preview
+        }
+
         let firstSection = Section()
         form +++ firstSection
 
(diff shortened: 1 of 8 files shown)


Sources/App/WebView/WebViewController.swift
```diff
@@ -833,6 +833,18 @@ final class WebViewController: UIViewController, WKNavigationDelegate, WKUIDeleg
             }
         )
     }
+
+    public func openActionAutomationEditor(actionId: String) {
```
zacwest, line 837: if this requires a certain core version, you should gate the whole experience on that version

Sources/App/Settings/SettingsDetailViewController.swift
```diff
@@ -737,6 +737,16 @@ class SettingsDetailViewController: HAFormViewController, TypedRowControllerType
                     }.done {
                         self?.updatePositions()
                     }.cauterize()
+
+                    if vc.shouldOpenAutomationEditor {
+                        vc.navigationController?.dismiss(animated: true, completion: {
+                            Current.sceneManager.webViewWindowControllerPromise.done { controller in
+                                controller.webViewControllerPromise.done { webViewController in
+                                    webViewController.openActionAutomationEditor(actionId: vc.action.ID)
```
zacwest: `then` lets you avoid this double layer of done chaining, e.g. `thing1.then { thing2 }.done { … }` and i think `.then(\.webViewControllerPromise)` may be all you need

Sources/App/Settings/ActionConfigurator.swift
```diff
@@ -343,20 +334,7 @@ class ActionPreview: UIView {
...
-        layer.masksToBounds = true
-
-        layer.shadowColor = UIColor.black.cgColor
-        layer.shadowOffset = CGSize(width: 0, height: 2.0)
-        layer.shadowRadius = 2.0
-        layer.shadowOpacity = 0.5
-        layer.masksToBounds = false
-        layer.shadowPath = UIBezierPath(roundedRect: bounds, cornerRadius: layer.cornerRadius).cgPath
-
+        layer.cornerRadius = 8
```
zacwest, line 337: this is to match the (old) widget. if you want to make changes, go for it, but maybe update that or make it match the new widgets if you can

Sources/App/Settings/Eureka/YamlSection.swift
```diff
@@ -10,6 +10,10 @@ public final class YamlSection: Section {
 
         row.cellSetup { cell, _ in
             cell.textView.configureCodeFont()
+            cell.textView.isEditable = false
+        }
+        row.cellUpdate { cell, _ in
+            cell.textView.isEditable = false
```
zacwest: can you make this a property on the Section itself?

Sources/App/Settings/ActionConfigurator.swift
bgoncal: I don't think it would make sense to update a feature that iOS itself is already deprecating, I would focus on Apple Watch and new Widgets from now on, no need to remove the old "Today's view widget", but also not needed to take it in consideration when moving forward.

They already don't look that good
Screenshot 2024-01-25 at 09 32 49

zacwest (approved):

PR #2544, opened 2024-01-25
Update CarPlay message while entities are getting filtered

bgoncal:


Sources/App/Resources/en.lproj/Localizable.strings
```diff
@@ -61,6 +61,7 @@
 "carPlay.navigation.tab.actions" = "Actions";
 "carPlay.navigation.tab.areas" = "Areas";
 "carPlay.navigation.tab.domains" = "Control";
+"carPlay.state.loading.title" = "Loading...";
```
zacwest, line 64: Use … - it looks nicer

PR #2567, opened 2024-02-01
Fix crash when closing the macos app from full screen

bgoncal:
## Summary

Strong reference was causing a crash when closing macos app from full screen.


Sources/App/Scenes/WindowSizeObserver.swift
```diff
@@ -1,16 +1,17 @@
 import Foundation
+import Shared
 import UIKit
 
 final class WindowSizeObserver: NSObject {
     @objc private(set) var observedScene: UIWindowScene?
-    private var observation: NSKeyValueObservation?
+    private weak var observation: NSKeyValueObservation?
```
zacwest: this feels a little weird. do we want to stop observation more concretely? does this token automatically invalidate on deinit? iirc it does not but just wanna double check

zacwest (approved):

Sources/App/Scenes/WindowSizeObserver.swift
bgoncal: I have updated the logic to explicit invalidate the observation, I noticed that the previous approach was not even saving the window size anymore so it was good to now have merged it yet

Sources/App/Scenes/WindowSizeObserver.swift
```diff
@@ -23,6 +24,11 @@ final class WindowSizeObserver: NSObject {
         }
         #endif
     }
+
+    public func stopObserving() {
+        observation?.invalidate()
```
zacwest, line 29: couldn't this go in deinit instead? i guess being explicit is fine here though

zacwest (approved):

Sources/App/Scenes/WindowSizeObserver.swift
bgoncal: It could, but since we call "start" I think it's easier to follow the implementation if we also have an explicit "stop"

PR #2581, opened 2024-02-16
Update dependecy and cleanup code

mat1th:
## Summary
This pr intruduces the following changes: 
- Upgrades the [Alomofire dependecy](https://github.com/home-assistant/iOS/commit/142efae0cced85bdc7baf0f7e50ba222402fcc65).
- Removes not needed  [available anotations](https://github.com/home-assistant/iOS/commit/a4e71a3c88cc008b98d785a524933a80e77cf3df) and split up code in multiple files; The anotations are not needed since https://github.com/home-assistant/iOS/pull/2469.

## Screenshots
No user faceing changes

NA

## Any other notes
NA


bgoncal:
I'll test these changes next week, thank you for contributing!

Sources/Shared/API/MJPEGStreamer.swift
bgoncal, line 1: No need to call super.urlSession...? Could you check that?

mat1th: Good point, super is implementred, so calling it now;

Sources/Shared/API/MJPEGStreamer.swift
```diff
@@ -7,12 +7,13 @@ class MJPEGStreamerSessionDelegate: SessionDelegate {
     static var taskUserInfoKey: AnyHashable = "taskUserInfoKey"
 
     // if/when alamofire also implements this again, we need to update to handle it as the breakpoint between images
-    func urlSession(
+    override func urlSession(
```
bgoncal, line 10: @zacwest Based on the comment on this function, do you see a problem with this change? What needs to be checked?

zacwest: Does this still work? I think it should probably still be fine with the default implementation happening, but it's definitely necessary to manually test.

bgoncal: I just tested, it's working

bgoncal (approved):
LGTM! Just take a look at the conflicts and then we can merge it

PR #2596, opened 2024-02-20
Allow custom headers to be added

sevimuelli:
## Summary

This allows to add custom headers to the server settings, for example for Cloudflare to be able to authenticate. It works on a simple idea that all the services used supports to add custom headers. Basically in every place where a request is made, if it is made to the correct address, the headers are added. If needed, it checks the requested URL to the specific server URL as you don't want to send your secrets to some third party service.

This PR requires home-assistant/HAKit#47

So far everything works as expected. There are some open points:

- [ ] Add translations
- [ ] Add/Modify tests and run all tests
- [ ] Modify ImageAttachmentViewController to support custom headers. This is the only element which does not yet support to add headers

## Screenshots

customheaders

Documentation: home-assistant/companion.home-assistant#1035

## Any other notes

There are two issues I struggle with:
- How do I correctly add new translations? Right now everything is hardcoded. I checked out Lokalise but there I can't add something new. Or can only admins do this? I guess entering everything manually doesn't feel right?
- When I debug the app, either in the simulator or on the phone, I can't notifications get to work. But I need this to test the changes. I got an Apple Dev account and set it up as described in this project. Is this not possible in the debug version or am I doing something wrong? I think I read something somewhere but I can't find it anymore.


balloob:
Closing as upstream rejected

ideasman69:
is there a reason for rejecting this?

mixman68:
Why rejected this ?

bgoncal:
We are seaking for a more permanent/long term solution to guarantee security in the App, this PR, although it looks simple, adds one more special case for connecting to Home Assistant, which leads to more complexity when developing not just for iOS but mac, watch, extension (widgets, shortcuts, notifications, etc).

I understand the frustration, I can guarantee you we are working on this topic, but for now our supported approach is providing the app a public URL and an internal URL only.

PR #2606, opened 2024-02-26
Add thread credentials management screen

bgoncal:
## Screenshots

Screenshot 2024-02-26 at 15 54 17


Configuration/Entitlements/App-ios.entitlements
```diff
@@ -30,5 +30,7 @@
 	<array>
 		<string>$(AppIdentifierPrefix)$(BUNDLE_ID_PREFIX).HomeAssistant$(BUNDLE_ID_SUFFIX)</string>
 	</array>
+    <key>com.apple.developer.homekit</key>
```
bgoncal: This was added so we can prefill matter commissioning code without having to scan QR codes or such a thing, not very related to thread credentials but 100% something we will use.

PR #2616, opened 2024-02-29
Change `static var` to `static let` in `Version`

michalsrutek:
## Summary

Change `static var` to `static let` to prevent inadvertent change from somewhere in the codebase.


bgoncal:
Thank you!

PR #2642, opened 2024-03-14
`static var` to `static let`

michalsrutek:
## Summary

- `static let` plays nicely with modern Swift concurrency (async/await)
- `static let` makes it impossible to mutate the values - causing undefined behavior in some cases


HomeAssistant.xcodeproj/project.pbxproj
```diff
@@ -4834,7 +4834,7 @@
 			);
 			mainGroup = B657A8DD1CA646EB00121384;
 			packageReferences = (
-				427692E12B98B82500F24321 /* XCLocalSwiftPackageReference "Sources/PushServer/SharedPush" */,
+				427692E12B98B82500F24321 /* XCLocalSwiftPackageReference "SharedPush" */,
```
michalsrutek, line 4837: got these changes just by running the project 🤷

Sources/App/Scenes/WindowSizeObserver.swift
```diff
@@ -32,13 +32,13 @@ final class WindowSizeObserver: NSObject {
 }
 
 enum ScenesWindowSizeConfig {
-    static var defaultSceneLastestSystemFrameDataKey = "default-scene-latest-system-frame-data"
+    private static let defaultSceneLatestSystemFrameDataKey = "default-scene-latest-system-frame-data"
```
michalsrutek, line 35: also a typo - `Lastest` => `Latest`

Sources/Shared/Environment/Constants.swift
```diff
@@ -138,7 +138,6 @@ public extension Version {
     static let canSendDeviceID: Version = .init(minor: 104)
     static let pedometerIconsAvailable: Version = .init(minor: 105)
     static let tagWebhookAvailable: Version = .init(minor: 114, prerelease: "b5")
-    static let tagPlatformTrigger: Version = .init(minor: 115, prerelease: "any0")
```
michalsrutek, line 141: unused 🚮

bgoncal:
@zacwest I see some var becoming let inside structs, thats what I also default using even though struct behavior already assures it would be immutable, I know you have an opinion about this, wanna jump in?

Sources/Shared/Notifications/LocalPush/PushProviderConfiguration.swift
```diff
@@ -1,8 +1,8 @@
 public struct PushProviderConfiguration: Codable {
-    public static var providerConfigurationKey = "ha_configurations"
+    public static let providerConfigurationKey = "ha_configurations"
 
-    public var serverIdentifier: Identifier<Server>
-    public var settingsKey: String
```
zacwest, line 5: I think that cases like this - where it's a data type - shouldn't force the non-static members to be `let` - that's already done by the mutability state of the struct. It's not _in and of itself wrong_ to copy a struct and mutate the value for some reason, as long as the values are valid state changes -- which the type can guard if it wants.

zacwest (commented):
I think for the most part the changes seem good. I don't think the 'static var…' to 'static let…' cases are a problem.

Keep in mind: this project uses the 'World' pattern from Pointfree, so a lot of mutability is actually meant to be in `Current` rather than elsewhere, and that's the point where things should be overridden. So for example, `static let shared xyz = …` should actually _generally_ be in Current rather than a singleton on the type. Makes testing easier.

Sources/Shared/Notifications/LocalPush/PushProviderConfiguration.swift
bgoncal: Since there is no real impact now, I will merge this PR as-is, we can always change it to var on demand of course when needed to actually do something with the struct.

bgoncal:
Thank you!

PR #2646, opened 2024-03-14
iOS platform 15 for download_localized_strings.yml

michalsrutek:
## Summary

As the min target is [iOS 15](https://github.com/home-assistant/iOS/pull/2469), this just updates the platform version that gets used in the `Download Localized Strings` workflow.


bgoncal:
Thank you!

PR #2654, opened 2024-03-19
Remove `settingsButton` optionality

michalsrutek:
## Summary
As the closure always return a `UIButton`, the property itself does not need to be an (force-unwrapped) Optional.


bgoncal:
Thank you!

PR #2658, opened 2024-03-19
Use a `switch` for `CMMotionActivityConfidence.description`

michalsrutek:
## Summary
- use a `switch` instead of if-else if-else
- use type inference for better readability


bgoncal:
Thank you!

PR #2689, opened 2024-04-02
Add BuildMaterialDesignIconsFont to codegen build phases

bgoncal:


zacwest:
How expensive is the script execution when it noops? I think I made it fast but I don't remember.

bgoncal:
> How expensive is the script execution when it noops? I think I made it fast but I don't remember. 
> 

I need to check, something weird happened in the pipelines, nothing related to the script was merged but the material design .json file was not in the runner machine anymore... then I found this script but I couldn't find where it was being triggered before, was it manually only?

zacwest:
Podfile executes it

bgoncal:
> Podfile executes it

Ok, good heads up then, I will check why pod install is not generating those files in the runner machine

Issue #2462, opened 2023-11-29
`clear_notification` message will not update badge value over cellular (outside local push?)


jamieshaw:
**iOS device model, version and app version**
Model Name: iPhone 15 Pro Max
Software Version: iOS 17.1.1
App version: 2023.7 (2023.471)

**Home Assistant Core Version**
Home Assistant Core 2023.10.5
(Frontend 20231005)

**Describe the bug**
When pushing a `badge` value as part of a `clear_notification` message, if connected to Wi-Fi and using local push, the notification is dismissed and the badge value updated.
When connected to cellular, the notification is cleared but the badge value remains unchanged (regardless if 0 or above).

**To Reproduce**
```yaml
service: notify.jamie
data:
    title: 'Home Assistant'
    message: 'clear_notification'
    data:
        push:
            badge: 99
```

**Expected behavior**
Consistent behaviour either setting the value or not — preferably, `clear_notification` will allow badge updates silently in the background.
(I'm aware of `delete_alert`, which is similar, but different. It does not clear the notification unless setting the badge to 0 — clearing _all_ notifications).

**Screenshots**
https://www.dropbox.com/scl/fi/5qb47a3xjjfwfdma4cnke/home_assistant_notification_bug.MP4?rlkey=h2lcwj63invrub7jmktlt3jzs&dl=0
(Video showing push over Wi-Fi, then failing over cellular, and the successfully back on Wi-Fi)

**Additional context**
In the above video `notify.jamie` is a group notifier that links 3 `mobile_app` notifiers.

jamieshaw:
So as of Home Assistant Companion (for iOS) 2023.12.1 (2023.508), this appears to have been resolved.  I can no longer get badges of any value to display regardless of connection type.

Unfortunately, not the desired outcome of allowing badging of `clear_notification` messages, but consistent now, nonetheless.

jamieshaw:
So as of Home Assistant Companion (for iOS) 2023.12.1 (2023.508), this appears to have been resolved.  I can no longer get badges of any value to display regardless of connection type.

Unfortunately, not the desired outcome of allowing badging of `clear_notification` messages, but consistent now, nonetheless.

jamieshaw:
I take it back — having some weird notification issues where dynamic actions where not populating — restarting the phone _fixed_ dynamic actions but reintroduced the same bug here the badge value is not updated over cellular for `clear_notification`.

lesgrebe:
[...]

PR #2715, opened 2024-04-12
Update badge manually for clear_notification

Cyberbeni:
## Summary

I had the same issue as #2462 where the badge isn't updating every time when sending badge update with the clear_notification command, mostly when I'm not getting it through local push but sometimes with local push too. Since I'm not in the developer team, I can't test push related stuff. (It might also need to be released to testflight in order to be tested if HA is not prepared to handle using the sandbox push environment for development builds)

AFAIK, the badge should be handled by the OS. But since that is not happening, we should try setting it manually.

## Any other notes

I haven't noticed it not clearing notifications like someone mentioned in that issue but the increased delay should resolve that.

### DIFF
--- a/Sources/Shared/Notifications/NotificationCommands/NotificationsCommandManager.swift
+++ b/Sources/Shared/Notifications/NotificationCommands/NotificationsCommandManager.swift
@@ -86,7 +86,7 @@ private struct HandlerClearNotification: NotificationCommandHandler {
             UNUserNotificationCenter.current().removeDeliveredNotifications(withIdentifiers: keys)
         }
         if let aps = payload["aps"] as? [String: Any],
-           let badge = payload["badge"] as? Int
+           let badge = aps["badge"] as? Int
         {
             UIApplication.shared.applicationIconBadgeNumber = badge
         }
--- a/Sources/Shared/Notifications/NotificationCommands/NotificationsCommandManager.swift
+++ b/Sources/Shared/Notifications/NotificationCommands/NotificationsCommandManager.swift
@@ -85,9 +85,14 @@ private struct HandlerClearNotification: NotificationCommandHandler {
         if !keys.isEmpty {
             UNUserNotificationCenter.current().removeDeliveredNotifications(withIdentifiers: keys)
         }
+        if let aps = payload["aps"] as? [String: Any],
+           let badge = payload["badge"] as? Int
+        {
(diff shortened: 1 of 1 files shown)


bgoncal:
According to the docs now you are mixing `clear_badge` with `clear_notification` aren't you? https://companion.home-assistant.io/docs/notifications/notification-commands

Sources/Shared/Notifications/NotificationCommands/NotificationsCommandManager.swift
```diff
@@ -85,9 +85,14 @@ private struct HandlerClearNotification: NotificationCommandHandler {
...
         }
+        if let aps = payload["aps"] as? [String: Any],
+           let badge = aps["badge"] as? Int
+        {
+            UIApplication.shared.applicationIconBadgeNumber = badge
+        }
         // https://stackoverflow.com/a/56657888/6324550
         return Promise<Void> { seal in
-            DispatchQueue.main.async {
+            DispatchQueue.main.asyncAfter(deadline: .now() + 1) {
```
bgoncal, line 96: Why is the delay needed?

Cyberbeni:
https://companion.home-assistant.io/docs/notifications/notifications-basic/#badge

The documentation suggests sending a `delete_alert` command to only update the badge. Which suggests that all other commands also support updating the badge. It works (most of the times) with local push, so I'd expect most people who want to update badge while clearing the notification to do so by including the badge count with the `clear_notification` command.

Cyberbeni:
Ah, I looked at the logs in the iOS app and it seems like they are indeed legacy notifications and the badge count is not included 🙈 

```
2024-04-12 16:53:11.759 [Verbose] [main] [NotificationManager.swift:101] handleRemoteNotification(userInfo:) > remote notification: [AnyHashable("gcm.message_id"): 1712933561674129, AnyHashable("homeassistant"): {
    command = "clear_notification";
    tag = apnstest;
}, AnyHashable("google.c.fid"): d1ivFK0OYkiXu5kXVL0u5K, AnyHashable("webhook_id"): 7f21d1c32ee7bc0f560efd5c2224fae5ef18366a912820c44e87fb4e3d26fb25, AnyHashable("google.c.a.e"): 1, AnyHashable("google.c.a.m_l"): legacyNotification, AnyHashable("google.c.sender.id"): 331041709873, AnyHashable("aps"): {
    "content-available" = 1;
}]
```

PR #2749, opened 2024-04-29
Keep only necessary webviews alive in the app

bgoncal:
## Summary

I noticed we are losing web views during the usage of the App, for example, when the user logs out we create a new web view after it logs in, but the old webview was kept in memory due to strong reference of WKScriptMessageHandler or due to not be listed in our cached webviews dictionary (and then not removed).

This PR gains controls of creation and usage of webviews aswell as removing them when not needed anymore, it also fixes the issue that some users experience "failure to authenticate" notifications (that were partially caused by these ghost webviews)

### DIFF
--- a/Sources/App/WebView/IncomingURLHandler.swift
+++ b/Sources/App/WebView/IncomingURLHandler.swift
@@ -5,7 +5,8 @@ import SafariServices
 import Shared
 
 class IncomingURLHandler {
-    weak var windowController: WebViewWindowController?
+    private(set) weak var windowController: WebViewWindowController!
+
     init(windowController: WebViewWindowController) {
         self.windowController = windowController
         registerCallbackURLKitHandlers()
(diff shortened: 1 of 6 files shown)


bgoncal:
I still have to test on macOS, for iOS it is working great now, no webview lost in the way

bgoncal:
> I still have to test on macOS, for iOS it is working great now, no webview lost in the way

Test done, in mac the scenario was even worse, everytime a window was opened then closed, that webview lived in memory until the app is killed

Sources/App/WebView/IncomingURLHandler.swift
```diff
@@ -5,7 +5,7 @@ import SafariServices
 import Shared
 
 class IncomingURLHandler {
-    let windowController: WebViewWindowController
+    weak var windowController: WebViewWindowController?
```
zacwest:
```suggestion
    private(set) weak var windowController: WebViewWindowController?
```

zacwest: Given some of the cases below crash, might as well just make this a `!` instead of `?`

Sources/App/WebView/WebViewWindowController.swift
```diff
@@ -75,7 +75,7 @@ class WebViewWindowController {
                 // not changing anything, but handle the promises
                 updateRootViewController(to: rootController)
             } else {
-                if let webViewController = WebViewController(restoring: .init(restorationActivity)) {
+                if let webViewController = makeWebViewIfNotInCache(restorationType: .init(restorationActivity)) {
```
zacwest, line 78: How could this code path hit a case where it's cached already?

bgoncal: The idea of the `makeWebViewIfNotInCache` is to make sure all webviews will be tracked in the cached dictionary, so in this case if it is not cached already it will then make a new webview and cache, otherwise we don't have how to control webviews that are created and left behind.

Issue #2752, opened 2024-04-30
Status bar color follows wrong color variable


Nezz:
**iOS device model, version and app version**

Model Name:  iPhone 15 Pro
Software Version: iOS 17.4.1
App version: 2024.624

**Home Assistant Core Version**
2024.4.4, 2024.5.0-b4

**Describe the bug**
The color of the status bar follows the `--app-header-background-color` variable. This is inconsistent with the browser where it follows `--app-theme-color`. It'd be nice to use that variable if it's set.

**To Reproduce**
Create a theme which sets `app-theme-color`:
```
app-theme-color: "#88827B" # Average header color, used on the top of the browser
```

**Expected behavior**
`--app-theme-color` should be used if it's set. The header color could still be used as a fallback.

**Screenshots**

**Additional context**
The frontend sets the theme color here:
https://github.com/home-assistant/frontend/blob/7120ad99b9bdd3d6dc648deb6ae5285d986f8744/src/state/themes-mixin.ts#L134

In the default theme the value of  `--app-header-background-color` and `--app-theme-color` is the same.

bgoncal:
I had to revert the PR, the app theme color doesn't always match the navigation bar, which makes it look weird in some situations like dark mode using the default themes

Nezz:
@bgoncal I think that should be fixed by this:
https://github.com/home-assistant/frontend/pull/20671

bgoncal:
> Looks like the fix doesn't work. I'll try to get it patched up.

Let's wait a bit until this ir tested in the frontend and we are sure it will behave as expected, but I will keep this on the radar, feel free to drop a message in a week or two

Nezz:
Sounds good. I opened the fix:
https://github.com/home-assistant/frontend/pull/20758

bgoncal:
Closing this issue since both frontend and mobile PR's were merged, let me know in case its not behaving as expected

Nezz:
Thank you. I grabbed the latest build from TestFlight. It's now the right color in light mode, but in dark mode the color does not change to the dark one.

PR #2757, opened 2024-05-01
Use same variable for status bar color as frontend

bgoncal:
## Summary

We were using `--app-header-background-color` for status bar in the app and frontend uses `--app-theme-color`, this PR changes that to keep consistency with frontend

### DIFF
--- a/Sources/App/Utilities/ThemeColors.swift
+++ b/Sources/App/Utilities/ThemeColors.swift
@@ -6,6 +6,7 @@ public struct ThemeColors: Codable {
     public enum Color: String, CaseIterable {
         // these are in WebSocketBridge.js in this repo (we inject it)
         case appHeaderBackgroundColor = "--app-header-background-color"
+        case appThemeColor = "--app-theme-color"
         case primaryBackgroundColor = "--primary-background-color"
         case textPrimaryColor = "--text-primary-color"
         case primaryColor = "--primary-color"
@@ -77,6 +78,7 @@ private extension ThemeColors.Color {
         case .primaryBackgroundColor: return UIColor(red: 0.98, green: 0.98, blue: 0.98, alpha: 1.0)
         case .primaryColor: return UIColor.white
         case .textPrimaryColor: return UIColor.white
+        case .appThemeColor: return UIColor(red: 0.01, green: 0.66, blue: 0.96, alpha: 1.0)
         }
     }
 }
--- a/Sources/App/Resources/WebSocketBridge.js
+++ b/Sources/App/Resources/WebSocketBridge.js
@@ -8,6 +8,7 @@ const notifyThemeColors = () => {
 
         [
             '--app-header-background-color',
+            '--app-theme-color',
             '--primary-background-color',
             '--text-primary-color',
             '--primary-color',
--- a/Sources/App/WebView/WebViewController.swift
+++ b/Sources/App/WebView/WebViewController.swift
@@ -344,12 +344,12 @@ final class WebViewController: UIViewController, WKNavigationDelegate, WKUIDeleg
         webView?.scrollView.backgroundColor = cachedColors[.primaryBackgroundColor]
 
         if let statusBarView = view.viewWithTag(111) {
-            statusBarView.backgroundColor = cachedColors[.appHeaderBackgroundColor]
+            statusBarView.backgroundColor = cachedColors[.appThemeColor]
         }
 
         refreshControl.tintColor = cachedColors[.primaryColor]
 
-        let headerBackgroundIsLight = cachedColors[.appHeaderBackgroundColor].isLight
+        let headerBackgroundIsLight = cachedColors[.appThemeColor].isLight
         underlyingPreferredStatusBarStyle = headerBackgroundIsLight ? .darkContent : .lightContent
 
         setNeedsStatusBarAppearanceUpdate()


zacwest:
The purpose of this view is to extend the background of the top header so it looks seamless - like a navigation controller. Is the header background color always the theme color or is this going to introduce discrepancies that look bad? What does this look like in the default themes?

bgoncal:
> The purpose of this view is to extend the background of the top header so it looks seamless - like a navigation controller. Is the header background color always the theme color or is this going to introduce discrepancies that look bad? What does this look like in the default themes?

Indeed, but browsers also do this "trick" (https://github.com/home-assistant/iOS/issues/2752), but for browsers the frontend uses app-theme-color instead of the app-header, so it makes sense to align I guess. @bramkragten can you confirm this?

For the default themes it looks the same with or without this change

Nezz:
Yes, it'd align this way. It does not make a difference for the default theme, but it does for ones that use transparency for `app-header-background-color` like my theme:
https://github.com/Nezz/homeassistant-visionos-theme

I believe @bgoncal uses a similar theme too since he implemented https://github.com/home-assistant/frontend/pull/20473

Sources/App/Utilities/ThemeColors.swift
```diff
@@ -6,6 +6,7 @@ public struct ThemeColors: Codable {
     public enum Color: String, CaseIterable {
         // these are in WebSocketBridge.js in this repo (we inject it)
         case appHeaderBackgroundColor = "--app-header-background-color"
```
Nezz, line 8: I wonder if we should fall back to this if the theme color is not set?

bgoncal:
I checked with front end and app-theme-color will be always set, so no need to fallback

bramkragten (approved):

jockebq:
Will this also resolve this issue?
https://github.com/home-assistant/frontend/issues/18277

Nezz:
No, that's a separate issue. I'll try to get it fixed.

jockebq:
> No, that's a separate issue. I'll try to get it fixed.

Great, really looking forward to it! What's the ETA from a fix like this to a release on iOS?

jockebq:
> > > No, that's a separate issue. I'll try to get it fixed.
> > 
> > 
> > Great, really looking forward to it! What's the ETA from a fix like this to a release on iOS?
> 
> If it gets merged before the last week of the month it probably will be in the month's release, otherwise the month after. (Unless it is something critical)

Sounds great, if I join the Testflight beta, will things like this get released sooner in the beta?

bgoncal:
Yes

Nezz:
That issue/feature looks like a frontend change to me. You might not even need an app update.

jockebq:
> > > No, that's a separate issue. I'll try to get it fixed.
> > 
> > 
> > Great, really looking forward to it! What's the ETA from a fix like this to a release on iOS?
> 
> If it gets merged before the last week of the month it probably will be in the month's release, otherwise the month after. (Unless it is something critical)

Got the TestFlight update which pushed today, but the fix isn't there right? Because I still have the same issue as before?

Thank you!

bgoncal:
> > > > No, that's a separate issue. I'll try to get it fixed.
> > > 
> > > 
> > > Great, really looking forward to it! What's the ETA from a fix like this to a release on iOS?
> > 
> > 
> > If it gets merged before the last week of the month it probably will be in the month's release, otherwise the month after. (Unless it is something critical)
> 
> Got the TestFlight update which pushed today, but the fix isn't there right? Because I still have the same issue as before?
> 
> Thank you!

Nope, the PR got reverted because the front end font use "app-theme-color" with the same color as the top navigation bar all the times, so it looks weird in the app.
The goal is to extend the navigation bar color below the status bar, currently the only way possible is through `--app-header-background-color`

jockebq:
> > > > > No, that's a separate issue. I'll try to get it fixed.
> > > > 
> > > > 
> > > > Great, really looking forward to it! What's the ETA from a fix like this to a release on iOS?
> > > 
> > > 
> > > If it gets merged before the last week of the month it probably will be in the month's release, otherwise the month after. (Unless it is something critical)
> > 
> > 
> > Got the TestFlight update which pushed today, but the fix isn't there right? Because I still have the same issue as before?
> > Thank you!
> 
> Nope, the PR got reverted because the front end font use "app-theme-color" with the same color as the top navigation bar all the times, so it looks weird in the app. The goal is to extend the navigation bar color below the status bar, currently the only way possible is through `--app-header-background-color`

Is `--app-header-background-color` something I can do myself or is this a fix you will work on?

bgoncal:
@jockebq `--app-header-background-color` is a variable that you can set in your theme and customize the header background.

jockebq:
> @jockebq `--app-header-background-color` is a variable that you can set in your theme and customize the header background.

Ah, that I have already set. What I am talking about is this issue: https://github.com/home-assistant/frontend/issues/18277
That on an iPhone 14 Pro looks very weird because the app itself does not begin on the top of the screen. So if you have a background set like in the picture it looks very bad.

bgoncal:
Ok, I got it now, this PR was never meant to fix that, this was the original issue: https://github.com/home-assistant/iOS/issues/2752

ncodee:
Wasn't this suppose to fix the issue? 
I'm still seeing a solid background under the status bar, even with the new background option introduced in 2024.6.

Nezz:
That was reverted in https://github.com/home-assistant/iOS/pull/2768. Sadly there is no context about why it was reverted.

The frontend was fixed to provide the correct color variables in Home Assistant 2024.6 (see https://github.com/home-assistant/iOS/issues/2752 and https://github.com/home-assistant/frontend/pull/20758). I guess now the fix can be re-applied @bgoncal?

bgoncal:
@Nezz https://github.com/home-assistant/iOS/pull/2826

Issue #2655, opened 2024-03-19
Remove deprecated iOS 12 things


michalsrutek:
There are still things in the project annotated with `@available(iOS, deprecated: 13.0)`. Namely
- `UIWindow`'s `convenience init(haForiOS12: ())`
- `StateRestorationKey`
-  `viewController(withRestorationIdentifierPath identifierComponents: [String])`
- `SceneManagerPreSceneCompatibility`

which could be removed, as the min iOS version is now 15.

bgoncal:
Hey @michalsrutek, I see you started contributing to the project, could you help out with this one? No pressure though

bgoncal:
Thanks @mat1th 🥳

PR #2808, opened 2024-06-03
Remove some depricated code pointed in #2655;

mat1th:
## Summary
In issue #2655 there is some code referenced that is not needed anymore. So I've removed it.

## Screenshots
The app still launches on a iPhone simulator. Could not test the macOS target.

## Any other notes
-


Sources/App/Notifications/NotificationManager.swift
```diff
@@ -293,7 +293,7 @@ extension NotificationManager: UNUserNotificationCenterDelegate {
             return
         }
 
-        var methods: UNNotificationPresentationOptions = [.alert, .badge, .sound]
```
mat1th, line 296: alert is deprecated since iOS 14 see https://developer.apple.com/documentation/usernotifications/unnotificationpresentationoptions/alert

Sources/App/WebView/IncomingURLHandler.swift
```diff
@@ -104,7 +104,7 @@ class IncomingURLHandler {
                         pipeline: pipeline?.identifier ?? "",
                         autoStartRecording: autoStartRecording
                     )
-                case let .rejected(error):
+                case .rejected:
```
mat1th: Error value was not used;

bgoncal:
Thank you!
For the linter:
`bundle exec fastlane autocorrect`

Sources/App/WebView/IncomingURLHandler.swift
bgoncal: Perhaps this instead:
`Current.Log.error("Failed to obtain webview to open Assist In App: \(error.localizedDescription)")`

bgoncal (approved):
Looks good to me, I'll run some manual tests just to be sure

Sources/App/Notifications/NotificationManager.swift
bgoncal: Perhaps we should follow the deprecation message?
"
Deprecated
Use [list](https://developer.apple.com/documentation/usernotifications/unnotificationpresentationoptions/list) and [banner](https://developer.apple.com/documentation/usernotifications/unnotificationpresentationoptions/banner) instead.
"

Sources/App/Notifications/NotificationManager.swift
```diff
@@ -302,9 +302,6 @@ extension NotificationManager: UNUserNotificationCenterDelegate {
             if presentationOptions.contains("badge") {
                 methods.insert(.badge)
             }
-            if presentationOptions.contains("alert") {
-                methods.insert(.alert)
```
bgoncal, line 306: https://github.com/home-assistant/iOS/pull/2808/files/1a600675484cb4edafd57c13573d1c3b371ecf1c#r1631344720
Same here

bgoncal (changes requested):
Tested and all looks good, just this suggestion above , ler me know what you think

Sources/App/Notifications/NotificationManager.swift
mat1th: Good point, I've added them.

bgoncal (approved):

bgoncal:
Thank you!

PR #2858, opened 2024-07-19
Update widget UI to align with frontend

bgoncal:
## Screenshots

Screenshot 2024-07-19 at 13 38 35


Casey-Sheridan:
I'm not sure if this is the appropriate way to begin a discussion on this change, so please correct me if I'm wrong here.
Is there an opportunity to continue the background color element for the individual full size widgets? For those of us using colors to quickly identify the widget we want, changing the background to only black or only white is a little disappointing.

image

PR #2864, opened 2024-07-23
Remove old Today's view widget

bgoncal:
## Summary

Since all efforts are focused on WidgetKit ('new' iOS widgets) and all users from the supported versions of iOS can use the new widgets, I am removing this old Today's view widgets.


FeikoJoosten:
Bit late to bring up seeing as this is already in the latest release.

There's nothing about this removal in the apps change log. I had to find out from this removal by the companion app documentation.

Here are some things I've run into regarding the usage of the WidgetKit widgets.

* The new widget view is much larger in screen size. For example, I have 21 actions on my widgets view. With the new WidgetKit view, this takes up my phones (iphone 15 pro) entire screen. Where with the today view widget this was barely 1/3 of the screen space.

* Pressing an action with the new WidgetKit always makes my phone vibrate, and I can't find an option to disable this.

* The today view widget automatically grew in size based on the actions added through the configuration.yaml on the server, or through the manually configured actions through the companion app. The new WidgetKit widgets are limited to 10 actions max.

PR #2897, opened 2024-08-01
Add Apple Watch battery sensor

bgoncal:


TomBrien:
This seems to be rounding to nearest 5 %. Is that expected?

bgoncal:
> This seems to be rounding to nearest 5 %. Is that expected? 

It looks like the API returns changes every 5% change indeed.
(Unless I rounded it somehow and didn't notice, but that's the values that the API were outputting for me while debugging too)

bgoncal:
@TomBrien I just noticed that iPhone battery behaves the same

TomBrien:
Oh so it does. Never noticed that...

PR #2910, opened 2024-08-10
Add sharedEventDeviceInfo data into the "tag_scanned" event

n6ham:
sharedEventDeviceInfo contains "sourceDevicePermanentID",  "sourceDeviceName" and "sourceDeviceID" values. 

It is used in other action events (such as event triggered from the Apple Watch). "sourceDeviceName" can be used to send an actionable notification back to the device that was used to scan tag. Currently it's not possible, since these values are not dispatched in the event body.

### DIFF
--- a/Sources/Shared/API/HAAPI.swift
+++ b/Sources/Shared/API/HAAPI.swift
@@ -635,7 +635,7 @@ public class HomeAssistantAPI {
     public func tagEvent(
         tagPath: String
     ) -> (eventType: String, eventData: [String: String]) {
-        var eventData = [String: String]()
+        var eventData: [String: String] = sharedEventDeviceInfo
         eventData["tag_id"] = tagPath
         if server.info.version < .tagWebhookAvailable {
             eventData["device_id"] = Current.settingsStore.integrationDeviceID


bgoncal (approved):
LGTM!

n6ham:
@bgoncal thank you for the review! 
qq: is there a way to know when a binary containing this change would be pushed into the App Store?

bgoncal:
Usually in the last week of the month, thanks for your contribution!

n6ham:
@bgoncal I've updated the iOS app yesterday to the latest version 2024.8.1 (2024.791) and tried to capture the "tag_scanned" event form the iPhone. Weirdly enough - I don't see "sourceDevicePermanentID", "sourceDeviceName", "sourceDeviceID" in the event body.

Just want to double check, if this PR is included into the latest binary 2024.791?

My initial reaction was "something is missing in the sharedEventDeviceInfo() thus the empty fields are added but filtered filtered out later".

But I don't think "Constants.PermanentID" can be empty when the method is called for the "tag_scanned" event, but not empty when an action is triggered from the Apple Watch. 

```
    public var sharedEventDeviceInfo: [String: String] { [
        "sourceDevicePermanentID": Constants.PermanentID,
        "sourceDeviceName": server.info.setting(for: .overrideDeviceName) ?? Current.device.deviceName(),
        "sourceDeviceID": Current.settingsStore.deviceID,
    ] }

```

There should be "sourceDevicePermanentID" in the event, thus I suspect my changes somehow are not a part of this binary. Wdyt?

bgoncal:
Your changes are part of release 2024.8, have you tried to run and put some breakpoints? Glancing at what you shared I can't spot the issue.

Let me know after you try, if you can't find the solution I can check later in the week

Issue #2908, opened 2024-08-08
iPadOS Card rendering issue since HA 2024.8.0


PRProd:
**iOS device model, version and app version**
Model Name: iPad (7th generation)
Software Version: 17.6
App version: (2024.730)

**Home Assistant Core Version**
2024.8.0

**Describe the bug**
After upgrading Home Assistant core to 2024.8.0, the rendering of _some_ cards inside the iOS companion app has changed.  Rendering in Chrome, Firefox, Safari, and the Android app still behave as expected after the upgrade to HA 2024.8.0.

**To Reproduce**
 - Use HA Core 2024.8.0
 - Download [HA-Firemote](https://github.com/PRProd/HA-Firemote) available through HACS
 - Add one or more Firemote cards to your dashboard
<br>

**Expected behavior**
Rendering should look the same in the iPad companion app as it does everywhere else.
<br>

**Screenshots**

iPad Rendering through Safari

![image](https://github.com/user-attachments/assets/a34fbd1a-9d6e-4d86-82ca-9b76defdccb0)
<br>

iPad Rendering through Home Assistant iOS Companion App

![Image](https://github.com/user-attachments/assets/29907e3a-20b1-4372-b3c0-be2fa2d5e848)
<br>

**Additional context**
I have seen this behavior on a couple of other HACS cards as well, so I know it's not just limited to Firemote.  I am the author of the Firemote card, so if any card design changes need to be made, please let me know.

bgoncal:
`Companion App Settings >> Debugging >> Reset frontend cache` let me know if it helps, no change was done on the App side.

PRProd:
@bgoncal ~

Thank you for your reply.  I should have mentioned that I tried that multiple times before submitting the report.  I also restarted my iPad a few times too.

I just tried again.  I reset the frontend cache 4 times.  Stopped and restarted the iPad app.  Powered off, then powered on the iPad.  Pulled down on the dashboard to see the refresh throbber a frew times.  No change.

Since the iPad software hasn't changed on your end, I can only assume that something was changed in HA Core 2024.8.0 that somehow is affecting the way css, sizing, and spacing is being rendered by the iPad app.  That's just a guess on my part though.

PRProd:
This issue is not unique to me.  I just received this issue report from another Firemote user:
https://github.com/PRProd/HA-Firemote/issues/478

PRProd:
Another example of the rendering issue affecting a different card
[...]

PR #2913, opened 2024-08-11
Make user-agent headers accurate across device types

mjaksn:
## Summary

This change adds an explicit preferredContentMode setting to `WKWebViewConfiguration` to ensure the framework-generated user-agent header more accurately reflects the user's device type. Without this configuration value, requests from `WKWebView` on iPads running iPadOS 13 or later default to sending a user-agent string imitating a desktop Intel Mac.

Additionally, this change resolves https://github.com/home-assistant/iOS/issues/2908 regarding layout issues seen on some cards when running the app on iPadOS by preventing browser compatibility code from being incorrectly applied due to the user-agent value.


bgoncal:
Whats the relation between the user agent and the component rendering? This PR considers iPad but not macOS App

PR #2918, opened 2024-08-14
Use mobile web content for iPad and desktop for mac

bgoncal:

### DIFF
--- a/Sources/App/WebView/WebViewController.swift
+++ b/Sources/App/WebView/WebViewController.swift
@@ -136,6 +136,7 @@ final class WebViewController: UIViewController, WKNavigationDelegate, WKUIDeleg
 
         config.userContentController = userContentController
         config.applicationNameForUserAgent = HomeAssistantAPI.applicationNameForUserAgent
+        config.defaultWebpagePreferences.preferredContentMode = Current.isCatalyst ? .desktop : .mobile
 
         webView = WKWebView(frame: view!.frame, configuration: config)
         webView.isOpaque = false


bgoncal:
On hold while discussing #2908

Sources/App/WebView/WebViewController.swift
```diff
@@ -136,6 +136,7 @@ final class WebViewController: UIViewController, WKNavigationDelegate, WKUIDeleg
 
         config.userContentController = userContentController
         config.applicationNameForUserAgent = HomeAssistantAPI.applicationNameForUserAgent
+        config.defaultWebpagePreferences.preferredContentMode = Current.isCatalyst ? .desktop : .mobile
```
steverep, line 139: If desktop is still used for Mac, I think the resulting user agent string would still unfortunately result in always serving the legacy frontend build, correct?

bgoncal: If I set "mobile" to be used for Mac the result user agent is even worse:

`Mozilla/5.0 (iPad; CPU OS 14_5 like Mac OS X) AppleWebKit/605.1.15 (KHTML, like Gecko) Home Assistant/2024.8 (io.robbie.HomeAssistant.dev; build:2024; macOS 14.5.0) Mobile/HomeAssistant, like Safari`

Says "iPad" and "OS 14_5" which is not the iPad version (it's macOS version)

steverep: Ugh, yeah that's terrible.  Like I said, Apple's fault.

bgoncal: Indeed and looking at historic data, they are not working to improve this 😅

bgoncal:
Now we need a solution for macOS

PR #2934, opened 2024-08-24
New gauge widget variant; fix leaked continuation

nickeidler:
## Summary

Main changes:
- Add a new variant for "no min/max label" gauge widget. This allows for use cases where the user wants to label different gauges to differentiate them. For example: indoor and outdoor temperature.
- Fix a bug where continuation leak in `action` block was causing the widget to become unresponsive (both on my sim and local device). Note that I don't do much iOS development so please let me know if there is a more idiomatic / correct way to prevent this condition.

I decided to use a new gauge widget type instead of inferring the type from which config options were filled or empty because the `ParameterSummary` definition was becoming extremely unwieldy.

## Screenshots

| Current normal gauge | New gauge type without min/max | Config for new gauge type |
|--------|--------|--------|
| current normal gauge | normal without min or max | config for normal without min or max |


Sources/Extensions/AppIntents/Widget/Gauge/WidgetGaugeAppIntentTimelineProvider.swift
```diff
@@ -97,19 +98,30 @@ struct WidgetGaugeAppIntentTimelineProvider: AppIntentTimelineProvider {
...
         let action = await withCheckedContinuation { continuation in
+            var hasResumed = false
+
             configuration.action?.asAction { action in
+                guard !hasResumed else { return }
+                hasResumed = true
                 continuation.resume(returning: action)
             }
+
+            // Fallback: Ensure continuation is resumed in case `asAction` doesn't call the closure in time
```
bgoncal: I think this extra logic does not make sense given the `withCheckedContinuation` `await` already waits for the `asAction`.
What you could do is checking if configuration.action is not nil before moving forward

Sources/Extensions/AppIntents/Widget/Gauge/WidgetGaugeAppIntent.swift
```diff
@@ -136,6 +182,7 @@ struct WidgetGaugeAppIntent: WidgetConfigurationIntent {
 @available(iOS 17.0, macOS 14.0, watchOS 10.0, *)
 enum GaugeTypeAppEnum: String, Codable, Sendable, AppEnum {
     case normal
+    case withoutMinMax
```
bgoncal: Can you rename it to `singleLabel`?
I think this way it's clearer, one type shouldn't impact the name of the following one

Sources/Extensions/AppIntents/Widget/Gauge/WidgetGaugeAppIntent.swift
```diff
@@ -146,6 +193,10 @@ enum GaugeTypeAppEnum: String, Codable, Sendable, AppEnum {
             "widgets.gauge.parameters.gauge_type.normal",
             defaultValue: "Normal"
         )),
+        .withoutMinMax: DisplayRepresentation(title: .init(
+            "widgets.gauge.parameters.gauge_type.withoutMinMax",
```
bgoncal: After updating the enum, also update this key

bgoncal (changes requested):

Sources/Extensions/AppIntents/Widget/Gauge/WidgetGaugeAppIntentTimelineProvider.swift
bgoncal: I have fixed this in a separate PR, after I merge it, feel free to rebase:
https://github.com/home-assistant/iOS/pull/2942

Sources/Extensions/AppIntents/Widget/Gauge/WidgetGaugeAppIntent.swift
nickeidler: yup makes sense. lmk if you want to change the display name in the UI as well

Sources/Extensions/AppIntents/Widget/Gauge/WidgetGaugeAppIntentTimelineProvider.swift
nickeidler: rebased

Sources/Extensions/AppIntents/Widget/Gauge/WidgetGaugeAppIntent.swift
bgoncal: Perhaps can you add a comment for each one to describe what they are meant to? I was reading again the code and for someone that will jump to this file in the future, there are so many "labels". So... something like:

// Label X is meant for this
// Label Y is for that

nickeidler: added

bgoncal (approved):

bgoncal:
Thank you very much! This change will be added to v2024.9 and before that a testflight build will be available

PR #2945, opened 2024-08-27
add charging state sensor for Apple Watch

yyolk:
## Summary

Following the thread at https://github.com/home-assistant/iOS/discussions/1764#discussioncomment-10434403 , this is my novice attempt at adding https://developer.apple.com/documentation/watchkit/wkinterfacedevicebatterystate or the Watch Battery State, using #2897 as a reference.

## Screenshots

I don't know how to take a screenshot of the sensor reporting through the simulator, I can add them if pointed in the right direction.

> _I will look into this. I will look for the corresponding watch battery change docs if they exist and extend them._

## Any other notes

This is a rough pass. I used #2897 as a launching point - I read it carefully as this is my first ever code contribution for Swift / iOS, WatchOS project but I think I had enough heuristics to manage to _possibly_ get it in the vicinity of it being correct.

I also used `Sources/Shared/API/Webhook/Sensors/BatterySensor.swift` for including an additional `Battery State` sensor, while exposing both sensors similarly in name like it's another device. 

I think it could still be improved by using `Sources/Shared/API/Webhook/Sensors/BatterySensor.swift` as a reference. I tried to change as little as possible.

I wasn't sure if I needed to add anything to `Sources/App/WatchCommunicatorService.swift`, considering the commit in #2897, https://github.com/home-assistant/iOS/commit/f961d689e56355c5bdf16272013aab3633087272#diff-daa127042af3bc5719e4eb4e02e3c50bfcef50084f554e53475f4a4238fdd096

- I will lean heavily on comments and will iterate where requested, if the changes aren't edited directly by the maintainers. @bgoncal 
- I **do** think I ran the Unit Tests correctly and added to the one that was also changed in #2897


Sources/Shared/API/Webhook/Sensors/WatchBatterySensor.swift
```diff
@@ -15,16 +15,24 @@ final class WatchBatterySensor: SensorProvider {
...
+            let batteryState = Communicator.shared.mostRecentlyReceievedContext
+                .content[WatchContext.watchBatteryState.rawValue] as? DeviceBattery.State ?? .unplugged
             var battery = batteryDecimal > -1 ? Int(batteryDecimal * 100) : -1
-            let icon: String = BatteryIcon.forBatteryLevel(battery, state: .unplugged)
+            let icon: String = BatteryIcon.forBatteryLevel(battery, state: batteryState)
             sensors.append(WebhookSensor(
-                name: "Watch Battery",
-                uniqueID: "watch-battery",
+                name: "Watch Battery Level",
+                uniqueID: "watch-battery_level",
```
bgoncal: Since we already released this to prod, changing this ID may cause issues for users, let's keep it "watch-battery"

Sources/Shared/API/Webhook/Sensors/WatchBatterySensor.swift
```diff
@@ -15,16 +15,24 @@ final class WatchBatterySensor: SensorProvider {
...
                 icon: icon,
                 deviceClass: .battery,
                 state: battery,
                 unit: "%"
             ))
+            sensors.append(WebhookSensor(
+                name: "Watch Battery State",
+                uniqueID: "watch-battery_state",
```
bgoncal: Let's use just dashes here, so:
```suggestion
                uniqueID: "watch-battery-state",
```

Sources/Shared/API/Webhook/Sensors/WatchBatterySensor.swift
bgoncal:
```suggestion
                uniqueID: "watch-battery",
```

bgoncal (changes requested):
Your PR looks good, it should be as simple as that to add the battery state indeed, my only concern is if this will be useful or not given the frequency of updates.
Were you able to use it for a day on your Apple Watch to evaluate that?

bgoncal (approved):

yyolk:
Applied suggestions, those all make sense!

> Your PR looks good, it should be as simple as that to add the battery state indeed, my only concern is if this will be useful or not given the frequency of updates. Were you able to use it for a day on your Apple Watch to evaluate that?

I'm going to figure out how to do this and report back. I think it should be straight forward enough under free provisioning, by just adding my device as a target in Xcode. 

I'm also interested in seeing how it works will make note of the update frequency.

With just the battery level now my automation triggers when `> 99%` as fully charged and announcing that when it is. I'll compare it to when I get the 'fully charged' notification for the paired phone.

bgoncal:
Let's merge it and then you can test via testflight, the worst that can happen is that we have to revert the change

bgoncal:
I'll make a new testflight build this week, keep an eye on it, 2024.9

yyolk:
> I'll make a new testflight build this week, keep an eye on it, 2024.9

Alright! I just signed up through test flight and will keep an eye on it.

PR #2950, opened 2024-08-28
Fix watch empty state and reposition reload button for watchOS 10

bgoncal:
## Screenshots

CleanShot 2024-08-29 at 01 18 41@2x

### DIFF
--- a/HomeAssistant.xcodeproj/project.pbxproj
+++ b/HomeAssistant.xcodeproj/project.pbxproj
@@ -3440,13 +3440,6 @@
 			path = Entities;
 			sourceTree = "<group>";
 		};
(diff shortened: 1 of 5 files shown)


rhyswaywood:
image

I'm curious if this will fix the error I had on first launch when I updated to 2024.9. Or do you think this is a separate issue?

bgoncal:
> image
> 
> I'm curious if this will fix the error I had on first launch when I updated to 2024.9. Or do you think this is a separate issue?
> 
> Force quitting the app on the watch and then re-adding my actions fixed this for me. 

This PR adds an empty state for when your watch has no configuration yet, this empty state replaces this error message you see, but you still need to open the iOS app and create a config for your Apple Watch

PR #2967, opened 2024-09-02
Define dark and tinted variants for some custom icons

bgoncal:


bgoncal:
Removed duplicate icons but tinted versions need to be separate files (even though they are the same), tried symlink and it does not work either, xcode recreates asset file.

PR #3051, opened 2024-09-25
Detail widget homescreen

Penait1:
## Summary
Add the option to add the detail widget to your homescreen. It was available in the lock screen already but I would love to have the option to add it to my homescreen to replace current widgets I have. In the future I would like add a widget where you could display multiple values with icons in a more table style fashion

## Screenshots
image
image

## Any other notes
This is my first mr and my first written lines of Swift, this is a small mr to get my hands dirty.


bgoncal:
Functionality-wise looks good to me, we just need to improve the design, feel free to propose something, If I get some ideas I'll drop here as well

bgoncal:
`bundle exec fastlane autocorrect` btw to fix lint issues

Penait1:
Yes, I can only come up with adding a small home assistant logo in the top left or something. The problem of the current detail widget is it only supports a single stat with subtitles. It is just a lot of wasted space. I would like a widget with a key value table, which is better suited for the bigger space but would take me more time.

I’ll sleep about it and run the formatter and if you have a better idea let me know.

bgoncal:
IMG_D243480331E4-1

if you can achieve a design similar to that then I can approve this PR until we have a better approach

bgoncal (changes requested):
IMG_D243480331E4-1

if you can achieve a design similar to that then I can approve this PR until we have a better approach

Penait1:
I think I'll close this for now since its not an ideal solution and I would like to spend the/your time on reviewing the better approach which I think is more in the direction of https://github.com/home-assistant/iOS/pull/3067

PR #3060, opened 2024-09-26
Use extended PAN ID instead of mac extended addres to select thread network

bgoncal:

### DIFF
--- a/Sources/App/WebView/WebViewExternalMessageHandler.swift
+++ b/Sources/App/WebView/WebViewExternalMessageHandler.swift
@@ -253,11 +253,14 @@ final class WebViewExternalMessageHandler {
     private func matterComissioningHandler(incomingMessage: WebSocketMessage) {
         if let preferredNetWorkMacExtendedAddress = incomingMessage.Payload?["mac_extended_address"] as? String,
            let preferredNetWorkActiveOperationalDataset = incomingMessage
-           .Payload?["active_operational_dataset"] as? String {
+           .Payload?["active_operational_dataset"] as? String,
+           let preferredNetworkExtendedPANID = incomingMessage.Payload?["extended_pan_id"] as? String {
             Current.settingsStore
                 .matterLastPreferredNetWorkMacExtendedAddress = preferredNetWorkMacExtendedAddress
             Current.settingsStore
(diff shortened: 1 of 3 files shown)


Sources/Extensions/Matter/MatterRequestHandler.swift
```diff
@@ -32,12 +32,16 @@ class MatterRequestHandler: MatterAddDeviceExtensionRequestHandler {
             )
         Current.Log
             .verbose(
-                "threadScanResults: \(threadScanResults.map { "Network name: \($0.networkName), Mac extended address: \($0.extendedAddress.hexadecimal)" })"
+                "threadScanResults: \(threadScanResults.map { "Network name: \($0.networkName), Mac extended address: \($0.extendedAddress.hexadecimal), Extended PAN ID: \($0.extendedPANID)" })"
```
agners: Detail: I'd move the extended PAN ID next to the network name, as these are representing the network:

```suggestion
                "threadScanResults: \(threadScanResults.map { "Network name: \($0.networkName), Extended PAN ID: \($0.extendedPANID), Mac extended address: \($0.extendedAddress.hexadecimal)" })"
```

Sources/App/WebView/WebViewExternalMessageHandler.swift
```diff
@@ -253,11 +253,14 @@ final class WebViewExternalMessageHandler {
     private func matterComissioningHandler(incomingMessage: WebSocketMessage) {
         if let preferredNetWorkMacExtendedAddress = incomingMessage.Payload?["mac_extended_address"] as? String,
```
agners, line 254: So `mac_extended_address` is essentially not used for the commissioning then no? Should we maybe even drop it? This got added to the message recently no?

Since we discuss to drop tracking the preferred border router entirely, I'd prefer if we don't tie things to much on the extended address unnecessarily.

bgoncal: As discussed in discord, this is the flow and what it requires:

Flow: 
1. Frontend sends mac extended address, dataset and extended PAN ID 
2. App uses mac extended address and dataset to save thread credential in keychain 
3. App moves on with matter flow 
4. When app gets thread scan results it uses the extended PAN ID to select the network

bgoncal:
Depends on frontend change: https://github.com/home-assistant/frontend/pull/22099

Issue #3076, opened 2024-10-03
"Enter Adress Manually" doesn't work


Moritz-Schmidt:
**iOS device model, version and app version**

Model Name: iPad Air (5th Generation) & iPhone 12 Pro
Software Version: 18.0 (iPad) & 17.6 (iPhone)
App version: 2024.9.3

**Home Assistant Core Version**
2024.10.0 (this should be irrelevant for the Issue)

**Describe the bug**
While setting up the companion app you have to press the "Continue" button multiple times, this is inconvenient but not the main problem.

On the "Scanning for Servers" page nothing happens when pressing the "Enter Address Manually" button.

**To Reproduce**
Open freshly installed Companion App, click on "continue", click on "Enter Adress Manually"

**Expected behavior**
Being able to enter the Adress manually.

karlbeecken:
Can reproduce, have the same issue

aleks-mariusz:
ditto, just installed on ipad and stuck at 'Scanning for servers' and can't hit that button to manually enter address

bgoncal:
I'm on it

bgoncal:
Can one of you share a screen recording? On my iPhone it is working fine, I'll check iPad in a few hours

Penait1:
I can reproduce it on my simulator. It throws the error
`[Error] [main] [OnboardingNavigationViewController.swift:129] show(_:sender:) > unknown sender Optional(<HomeAssistant.OnboardingScanningViewController: 0x108f12610>) wanted us to present: <HomeAssistant.OnboardingManualURLViewController: 0x129808a00>`

I'm not sure why this condition exists near line 129 
`else if sender as? UIViewController == topViewController`

Adding an explicit else caluse for the OnboardingManualURLViewController fixes the issue but I'm not sure thats best practise

bgoncal:
My PR above should handle that, I'll let you know it is available for test

rgrae81:
iPhone 12 mini iOS 17.6 and 17.7 same Problem.
iPhone 14 Pro iOS 17.6 is working fine.

bgoncal:
I just submitted beta with the fix for it to Apple Review, it will come as 2024.9.4

aleks-mariusz:
thanks for the lightning quick turnaround folks! curious how long the take it takes to show up in the app store for us mere mortals these days is?

CedricHaidacher:
[...]

PR #3079, opened 2024-10-03
Continue navigationlink match button size

Penait1:
## Summary
Make the NavigationLink match the size of the button. The button only triggers when pressing the button text, which makes it frustrating to press (See https://github.com/home-assistant/iOS/issues/3076 bug description)

This pr makes the clickable part of the button larger making it easier to press.


bgoncal:
Thanks dude!

bgoncal (approved):

PR #3098, opened 2024-10-24
Avoid confirmation when running legacy iOS Action from Apple Watch

bgoncal:


Sources/Extensions/Watch/Home/MagicItemRow/WatchMagicViewRowViewModel.swift
```diff
@@ -24,8 +24,7 @@ final class WatchMagicViewRowViewModel: ObservableObject {
     }
 
     func executeItem() {
-        guard state != .loading else { return }
```
bgoncal, line 27: This also enables tapping more than once on an item before waiting it to end (request from community)

PR #3067, opened 2024-09-28
Sensor State Widget

Penait1:
## Summary
Added a widget for the homescreen that fetches the current sensors in home assistant and displays its state. It's a very basic first version, I would like to get this moving to be able to expand upon it in the future like optional icons and able to alter the sensor name text so it takes up less space. Only the systemSmall size is supported for now

## Screenshots
Simulator Screenshot - iPhone 16 Pro - 2024-09-28 at 14 14 02
Simulator Screenshot - iPhone 16 Pro - 2024-09-28 at 14 14 28
Simulator Screenshot - iPhone 16 Pro - 2024-09-28 at 14 07 07

## Any other notes
It's maybe a rough MR, I'm not that experienced in Swift/iOS development so I need some pointers how to proceed further and like to keep it small. Its functional but for example all the translation strings are missing. Also extending the Domain with the enum value `.sensors` required changes in the CarPlay section which is not ideal. Also I'm not sure about the name of the widget, maybe it just should be called Sensors instead of `DetailTable`.

PR #3102, opened 2024-10-27
Group by server, widget selection order

Penait1:
## Summary
Fixed two issues mentioned in #3067;

Sensors in the suggested sensors view should be grouped by server now
The order of the selected sensors is now respected if they come from two different servers

## Screenshots
Screenshot 2024-10-27 at 08 06 20

## Any other notes
The `for` loop bothers me, Swift feels a lot like Kotlin and a `map` seems more appriopiate.
Because an async call happens in it I could not get `map` working that is nice to read, let me know if there is a better option chatgpt gave me this

```
let sensorValues = try await withThrowingTaskGroup(of: WidgetSensorsEntry.SensorData?.self) { group in
    for sensor in configuration.sensors ?? [] {
        group.addTask {
            guard let server = Current.servers.all.first(where: { $0.identifier.rawValue == sensor.serverId }) else {
                throw WidgetSensorsDataError.noServers
            }
            return try await fetchSensorData(for: sensor, server: server)
        }
    }

    return try await group.compactMap { $0 }
}
```

### DIFF
--- a/Sources/Extensions/AppIntents/Sensor/IntentSensorsAppEntity.swift
+++ b/Sources/Extensions/AppIntents/Sensor/IntentSensorsAppEntity.swift
@@ -38,8 +38,12 @@ struct IntentDetailsTableAppEntityQuery: EntityQuery {
         getSensorEntities().flatMap(\.value).filter { identifiers.contains($0.id) }
     }
 
-    func suggestedEntities() async throws -> [IntentSensorsAppEntity] {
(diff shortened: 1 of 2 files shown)


bgoncal:
The solution provided is fine. About what chatgpt suggested, you can read more about it here https://www.swiftbysundell.com/articles/async-and-concurrent-forEach-and-map/

bgoncal (approved):

PR #3105, opened 2024-10-28
Respect frontend "start_listening" value when launching Assist

bgoncal:


bgoncal:
@bramkragten with this change, the standard Assist button from the navigation bar now open Assist in iOS directly in voice mode, is this expected? FE is sending "start_listening: true". Just double checking because thats not the behavior on a browser

PR #3108, opened 2024-10-29
Preserve order getEntities function

Penait1:
## Summary
Changed the implementation of `getEntities(matching string: String? = nil)` to return an array of key values pairs instead of a dictioniary. The intent of the function seems to return a sorted list of entities grouped by server name since the servers are sorted `Current.servers.all.sorted(by: { $0.info.name < $1.info.name })`. A dictioniary does not guarantee order, resulting in unpredicting results in my `Sensor` widget configure window. Sometimes one server is first, the next time the other. 

Apple also recommends using a KeyValuePair when order is important:

https://developer.apple.com/documentation/swift/keyvaluepairs
`Use a KeyValuePairs instance when you need an ordered collection of key-value pairs and don’t require the fast key lookup that the Dictionary type provides.`

Because there are typically (I think?) not many servers configured the loss of fast lookup seems acceptable to me.

## Screenshots
Here an overview of the `getEntities` function while debugging showing the problem. I adjusted the function slightly to temp store the values, `servers` contains the value of `Current.servers.all.sorted(by: { $0.info.name < $1.info.name })` The entitiesPerSever does not have the same order as the `servers` variable, while they should have

Screenshot 2024-10-29 at 17 16 04

## Any other notes
This is also an issue for other widgets like the script widget. I only fixed the function in `IntentSensorsAppEntity` though
, they have the same problem in the calling function because they return an dictionary not preserving order;
`private func getScriptEntities(matching string: String? = nil) -> [Server: [IntentScriptEntity]] {`

### DIFF
--- a/Sources/Extensions/AppIntents/Script/ScriptAppIntent.swift
+++ b/Sources/Extensions/AppIntents/Script/ScriptAppIntent.swift
@@ -113,7 +113,7 @@ struct IntentScriptEntity: AppEntity {
 @available(iOS 16.4, macOS 13.0, watchOS 9.0, *)
 struct IntentScriptAppEntityQuery: EntityQuery, EntityStringQuery {
     func entities(for identifiers: [String]) async throws -> [IntentScriptEntity] {
-        getScriptEntities().flatMap(\.value).filter { identifiers.contains($0.id) }
+        getScriptEntities().flatMap(\.1).filter { identifiers.contains($0.id) }
     }
 
     func entities(matching string: String) async throws -> IntentItemCollection<IntentScriptEntity> {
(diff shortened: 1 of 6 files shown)


bgoncal:
Thats a good catch, but you are returning a tuples array not KeyValuePairs right?

Penait1:
Ah my bad that is true. I came across the KeyValuePair documentation next to someone posting this solution and thought it was the same.

I tried a KeyValuePair but it appears it is not mutable (for appends) so maybe a tuples array is the best for now

bgoncal:
Using KeyValuePair you wouldn't append, you need to set the value for a given key, can you try that? Otherwise tuple is fine of course.

I see you changed the `ControlEntityProvider`, have you checked if this doesn't break other widgets or iOS 18 control center controls?

Penait1:
Setting also does not seem to work.

I updated the mr and fixed other `AppEntity` classes that called the `getEntity` function. They seem to work fine and they also display the servers now in consistent order.

bgoncal (approved):

PR #3126, opened 2024-11-04
Change sensor widget view

Penait1:
## Summary
Changed the sensors widget to the `WidgetBasicView` as discussed with @bgoncal.
Also added support for tap widget to refresh
## Screenshots
Screenshot 2024-11-04 at 19 01 02


Sources/Extensions/Widgets/Common/WidgetBasicContainerView.swift
```diff
@@ -52,6 +52,8 @@ struct WidgetBasicContainerView: View {
                 intent.showConfirmationNotification = showConfirmationNotification
                 return intent
             }
+        case .refresh:
+            return ReloadWidgetsAppIntent()
```
bgoncal: Since this is also an AppIntent we don't need an extra case for that, can you reuse the ".appIntent(widgetIntentType)"?

Sources/Extensions/AppIntents/Widget/Sensor/WidgetSensorsAppIntentTimelineProvider.swift
```diff
@@ -167,10 +166,11 @@ struct WidgetSensorsEntry: TimelineEntry {
     var sensorData: [SensorData] = []
 
     struct SensorData {
-        var entityId: String
+        var id: String
```
bgoncal, line 169: is this the entityID or the "serverId-entityId"?

bgoncal (changes requested):
Some small tweaks, we may also need to change the design a little more but that can be in a separate PR

Sources/Extensions/AppIntents/Widget/Sensor/WidgetSensorsAppIntentTimelineProvider.swift
Penait1: You are right, I wanted to make it more uniform with the other widgets but forgot changing the content. I changed it so the value is filled with `sensor.id` instead of  `sensor.entityId` now

Sources/Extensions/Widgets/Common/WidgetBasicContainerView.swift
Penait1: I moved it from the `InteractionType` to the`WidgetIntentType` enum, I hope that is what you meant because I think I still need an extra case

bgoncal (approved):
As mentioned above, we will have a redesign at some point, but this is already progress compared to before, so I'll merge it

PR #3172, opened 2024-11-18
Redesign Sensor widget

Penait1:
## Summary
Redesign of the Sensor widget

## Screenshots
Screenshot 2024-11-20 at 17 04 05
Screenshot 2024-11-19 at 19 52 17


Penait1:
@bgoncal I think its ready for review, the test step seems to be failing but I'm not sure why

Sources/Extensions/Widgets/Common/WidgetBasicContainerView.swift
```diff
@@ -86,22 +88,26 @@ struct WidgetBasicContainerView: View {
...
                 HStack(spacing: Spaces.one) {
                     ForEach(column) { model in
                         if case let .widgetURL(url) = model.interactionType {
                             Link(destination: url.withWidgetAuthenticity()) {
                                 if #available(iOS 18.0, *) {
-                                    WidgetBasicViewTintedWrapper(model: model, sizeStyle: sizeStyle)
+                                    tintedWrapperView(model: model, sizeStyle: sizeStyle)
                                 } else {
-                                    WidgetBasicView(model: model, sizeStyle: sizeStyle, tinted: false)
+                                    WidgetBasicButtonView(model: model, sizeStyle: sizeStyle, tinted: false)
```
bgoncal: Should this case always be a button view?

bgoncal (changes requested):
Just tested on my device and it looks great! One minor thing, the widget background needs to be that gray-ish that the other widgets before, because we try tor eplicate the lovelace UI, gray-ish bg + white background tile, I think the design I shared with you had the wrong bg as well, my bad.

Also are you planning to add more sizes? (can be in a separate PR of course)

Sources/Extensions/Widgets/Common/WidgetBasicContainerView.swift
Penait1: Mm no I was mistaken

Penait1:
Yes I'm planning more sizes but I'll do it in a separate pr. 

I quickly checked if the sensor type view works with bigger sizes and it works. I'll have to do some extra stuff for the snapshot preview still I think

bgoncal (approved):

PR #3177, opened 2024-11-19
simplify for loop

Penait1:
## Summary
Really small mr, I just came across it looking into the scripts widget. I assume the for loop only returns when all sync operations are finished, removing the need to manually check for it

The preview still seems to work fine when adding a widget


bgoncal (approved):
Good catch, I think in the past there was an API call in the middle

PR #3483, opened 2025-03-17
Add support to Assist stream response

bgoncal:

### DIFF
--- a/Sources/App/WatchCommunicatorService.swift
+++ b/Sources/App/WatchCommunicatorService.swift
@@ -341,6 +341,10 @@ extension WatchCommunicatorService {
 // MARK: - AssistServiceDelegate
 
 extension WatchCommunicatorService: AssistServiceDelegate {
+    func didReceiveStreamResponseChunk(_ content: String) {
+        /* no-op */
+    }
+
     func didReceiveEvent(_ event: Shared.AssistEvent) {
         Current.Log.info("Watch Assist received event: \(event)")
     }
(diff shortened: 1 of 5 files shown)


Sources/App/WatchCommunicatorService.swift
```diff
@@ -341,6 +341,10 @@ extension WatchCommunicatorService {
 // MARK: - AssistServiceDelegate
 
 extension WatchCommunicatorService: AssistServiceDelegate {
+    func didReceiveStreamResponseChunk(_ content: String) {
+        /* no-op */
+    }
```
TimoPtr, line 346: I guess watch doesn't have text feedback and that's why we don't need to do anything?

Sources/Shared/Intents/AppIntent/AssistInApp/AssistService.swift
```diff
@@ -184,6 +186,8 @@ public final class AssistService: AssistServiceProtocol {
             guard let mediaUrlPath = data.data?.ttsOutput?.urlPath,
                   let mediaUrl = server.info.connection.activeURL()?.appendingPathComponent(mediaUrlPath) else { return }
             delegate?.didReceiveTtsMediaUrl(mediaUrl)
+        case .intentProgress:
+            delegate?.didReceiveStreamResponseChunk(data.data?.chatLogDelta?.content ?? "Unknown")
```
TimoPtr, line 190: Do you really want to display `Unknown` if the intent isn't containing any data?

bgoncal: It's just a visual feedback that the content wasn't available since we can't guarantee it due to the json changing depending on event type, on the other hand according to the docs it shouldn't happen that the content is empty or nil

Sources/App/WatchCommunicatorService.swift
bgoncal: The watch has text feedback, just not stream support yet, this will require more work

PR #3484, opened 2025-03-17
Snapshot preview reuse

banaslee:
## Summary
SnapshottablePreviews applied in the LocationHistoryListView and Widgets.

In LocationHistoryListView, due to how Realm is used and how previews are implemented under the hood, the injected data is written both on `init` and `onAppear`.

For Widgets, not all the previews are visible. This is a limitation of the previews UI. 

I also ended up removing the duplicate snapshots that would assert the behavior when more than the allowed tiles were passed. This can be restored if preferred.


bgoncal:
Can you split this into more than 1 PR? One for the sharing new target, one for location history and one for widgets.
Just to be easier to review, but if it endup being hard it's fine, I'll just take more time to review since for these changes I'll have to checkout the branch.
Github is not even loading the snapshot diffs to me

banaslee:
> Can you split this into more than 1 PR? One for the sharing new target, one for location history and one for widgets. Just to be easier to review, but if it endup being hard it's fine, I'll just take more time to review since for these changes I'll have to checkout the branch. Github is not even loading the snapshot diffs to me

I can. First let me know if this filter helps https://github.com/home-assistant/iOS/pull/3484/files?file-filters%5B%5D=.h&file-filters%5B%5D=.lock&file-filters%5B%5D=.pbxproj&file-filters%5B%5D=.swift&file-filters%5B%5D=.xcscheme&file-filters%5B%5D=No+extension&show-deleted-files=true&show-viewed-files=true

It shows you only the non-snapshot changes. 

Also, it's important to note that snapshots have not changed. You can see from the navigation on the left side of the screen (if on web) that they were either removed or renamed. The ones removed were the ones for unsupported number of tiles.

bgoncal:
It looks fine to me, but I would at least move the new framework-add to a separate PR, so it's easier to revert if needed and also separate purposes for future reference.

banaslee:
> It looks fine to me, but I would at least move the new framework-add to a separate PR, so it's easier to revert if needed and also separate purposes for future reference.

Will do. A PR adding the framework plus logic and another using the new tooling.

bgoncal (approved):

PR #3486, opened 2025-03-18
Implement Assist continous conversation

bgoncal:

### DIFF
--- a/Sources/Shared/Intents/AppIntent/AssistInApp/AssistService.swift
+++ b/Sources/Shared/Intents/AppIntent/AssistInApp/AssistService.swift
@@ -4,6 +4,7 @@ import HAKit
 
 public protocol AssistServiceProtocol {
     var delegate: AssistServiceDelegate? { get set }
+    var shouldStartListeningAgainAfterPlaybackEnd: Bool { get }
     func replaceServer(server: Server)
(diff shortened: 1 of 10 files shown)


Sources/Shared/Intents/AppIntent/AssistInApp/AssistService.swift
```diff
@@ -153,48 +155,73 @@ public final class AssistService: AssistServiceProtocol {
...
+    }
+
+    private func sttEnd(content: String?) {
+        delegate?.didReceiveSttContent(content.orEmpty)
+    }
+
+    private func intentEnd(conversationId: String?, content: String?, continueConversation: Bool) {
+        self.conversationId = conversationId
+        delegate?.didReceiveIntentEndContent(content.orEmpty)
+        shouldStartListeningAgainAfterPlaybackEnd = continueConversation
```
TimoPtr, line 208: On Android I did set it back to false once I started to listen again. But it might be useless since if you start listening again you should receive an INTENT-END with continueConversation to either true or false. Does it change something for you?

bgoncal: Good catch, but indeed, it will always be true or false, and if it is nil, it defaults to false

PR #3487, opened 2025-03-19
Create SharedTesting framework target

banaslee:
## Summary
Put together a technique to reuse previews to generate and assert snapshots.
This either saves time writing snapshot tests or makes snapshot tests more useful by using them as previews.

This is implemented in a couple of libraries already ([example](https://github.com/BarredEwe/Prefire)). Went with a simple implementation here but a library can be adopted to leverage additional features and support.

To hold the code for snapshotting, a new framework was introduced, SharedTesting, to collect all the utilities built for testing. It must only be imported in test targets.

The preview side of the code lives in the Shared framework.

Applying this new technique will come in a separate PR.

[Original PR](https://github.com/home-assistant/iOS/pull/3484)


bgoncal (approved):
LGTM

banaslee:
> LGTM

Cool. Let me know if you need something before merging.

bgoncal:
@banaslee Distribution pipeline is failing after merging this PR, can you check? Otherwise I can revert the merge for now
https://github.com/home-assistant/iOS/actions/runs/13965048702

bgoncal:
I'll revert and we trigger the distribution pipeline manually again before merging it once more

banaslee:
This is the failure @bgoncal 
`/Users/runner/work/iOS/iOS/HomeAssistant.xcodeproj: error: SharedTesting does not support provisioning profiles. SharedTesting does not support provisioning profiles, but provisioning profile iOS App Store - SharedTesting has been manually specified. Set the provisioning profile value to "Automatic" in the build settings editor. (in target 'SharedTesting' from project 'HomeAssistant')`

I find strange is that the tests targets are also being caught by `gym`. I think it's safe to exclude them.

Though I'll check the build settings for the new target.

bgoncal:
Maybe you need to check `HomeAssistant.xcconfig`

banaslee:
> Maybe you need to check `HomeAssistant.xcconfig`

There was nothing there but the build settings had different settings than other frameworks. Visually they looked the same but the raw value was not.
I'll push to my branch.

Any way, I think it would be good to try and filter out targets that are not meant to export from the `provisioning_profile_specifiers` in Fastfile.

Those are two parallel changes though.

banaslee:
Changes to build settings have been pushed to the original branch. I'll open a new PR.

PR #3527, opened 2025-04-07
Simplify and migrate onboarding screens to SwiftUI

bgoncal:

## Any other notes

https://github.com/user-attachments/assets/e79c6e4d-13ff-405a-9463-02e597ce4996


Sources/App/Onboarding/Container/OnboardingNavigationView.swift
```diff
@@ -0,0 +1,94 @@
...
+                case .initial:
+                    OnboardingWelcomeView(shouldDismissOnboarding: $viewModel.shouldDismiss)
+                case .secondary:
+                    OnboardingServersListView(shouldDismissOnboarding: $viewModel.shouldDismiss)
+                case let .required(type):
+                    switch type {
+                    case .full:
+                        OnboardingWelcomeView(shouldDismissOnboarding: $viewModel.shouldDismiss)
+                    case .permissions:
+                        OnboardingPermissionsNavigationView(onboardingServer: nil)
```
TimoPtr, line 60:
```suggestion
                        OnboardingPermissionsView(onboardingServer: nil)
```
To fit the other views. Except if I didn't get the pattern here.

Sources/App/Onboarding/Screens/OnboardingErrorView.swift
```diff
@@ -23,15 +23,20 @@ struct OnboardingErrorView: View {
             VStack {
                 exportLogsView
                 Button(L10n.Onboarding.ConnectionError.moreInfoButton) {
-                    openURL(documentationURL(for: error))
+                    if let url = documentationURL(for: error) {
+                        openURL(url)
+                    } else {
+                        Current.Log.error("Failed to create documentation URL for error: \(error)")
```
TimoPtr, line 29: So if this happen nothing is going to be displayed to the user?

Sources/App/Onboarding/Screens/OnboardingPermissions/LocationPermissionView.swift
```diff
@@ -0,0 +1,175 @@
...
+                .font(.title.bold())
+            Text(verbatim: L10n.Onboarding.Permission.Location.description)
+                .multilineTextAlignment(.center)
+                .opacity(0.5)
+            PrivacyNoteView(content: L10n.Onboarding.Permission.Location.privacyNote)
+        }
+        .frame(maxWidth: .infinity, alignment: .center)
+    }
+
+    private var bullets: some View {
```
TimoPtr: I don't think this is used?

Sources/App/Onboarding/Screens/OnboardingPermissions/LocationPermissionView.swift
```diff
@@ -0,0 +1,175 @@
...
+            Image(uiImage: permission.enableIcon.image(
+                ofSize: .init(width: 100, height: 100),
+                color: nil
+            ).withRenderingMode(.alwaysTemplate))
+                .foregroundStyle(Color.asset(Asset.Colors.haPrimary))
+            Text(verbatim: permission.title)
+                .font(.title.bold())
+            Text(verbatim: L10n.Onboarding.Permission.Location.description)
+                .multilineTextAlignment(.center)
+                .opacity(0.5)
```
TimoPtr, line 48: Side note: That's where maybe we should have a Design system because I think that we should have categories for txt so that we have color and size for subtitles and so on., instead of opacity I think we should have a color for all the subtitles.

Sources/App/Onboarding/Screens/OnboardingPermissions/LocationPermissionView.swift
```diff
@@ -0,0 +1,175 @@
...
+            .buttonStyle(.secondaryNegativeButton)
+        }
+    }
+}
+
+#Preview {
+    LocationPermissionView(permission: .location) {}
+}
+
+final class LocationPermissionViewModel: NSObject, ObservableObject {
```
TimoPtr: Is it common on iOS to keep this into the same file as the view? It seems that in some case you put it into a dedicated file and in other case it is in the same file.

Sources/App/Onboarding/Screens/OnboardingPermissions/LocationPermissionView.swift
```diff
@@ -0,0 +1,175 @@
...
+
+    func requestLocationPermission() {
+        locationManager.delegate = self
+        locationManager.requestWhenInUseAuthorization()
+    }
+
+    func disableLocationSensor() {
+        let sensorsToDisable = webhookSensors.filter { sensor in
+            sensorIdsToEnableDisable.map(\.rawValue).contains(sensor.UniqueID)
+        }
```
TimoPtr: You could extract this into a small private function to avoid duplicating it twice.

Sources/App/Onboarding/Container/OnboardingNavigationView.swift
TimoPtr: Ok I think I got it. It's for future proofing `OnboardingPermissionsNavigationView` should manage the navigation to ask for multiple permission but for now we only have Location. Is that right?

bgoncal: `OnboardingPermissionsNavigationView` is not a permission view per say, it is a navigation container for all required permission views, thats why the naming.

Sources/App/Onboarding/Screens/OnboardingPermissions/OnboardingPermissionsNavigationView.swift
```diff
@@ -0,0 +1,53 @@
...
+                    Current.onboardingObservation.complete()
+                }
+            } else {
+                flowEnd
+            }
+        }
+        .onDisappear {
+            Current.connectivity.syncNetworkInformation {
+                if let onboardingServer, let currentSSID = Current.connectivity.currentWiFiSSID() {
+                    // Update SSIDs if we have access to them, since we're gonna need it later
```
TimoPtr, line 30: Why do we care about SSID?

Sources/App/Onboarding/Screens/OnboardingServersList/ManualURLEntry/ManualURLEntryView.swift
```diff
@@ -0,0 +1,102 @@
...
+struct ManualURLEntryView: View {
+    @Environment(\.dismiss) private var dismiss
+    @State private var urlString = ""
+    @FocusState private var focused: Bool?
+    @State private var showInvalidURLError = false
+
+    let connectAction: (URL) -> Void
+
+    var body: some View {
+        NavigationView {
```
TimoPtr, line 13: If it contains a navigationView don't you want to reflect it in the name of the View like you did before?

Sources/App/Onboarding/Screens/OnboardingServersList/ManualURLEntry/ManualURLEntryView.swift
```diff
@@ -0,0 +1,102 @@
+import Shared
+import SwiftUI
+
+struct ManualURLEntryView: View {
+    @Environment(\.dismiss) private var dismiss
+    @State private var urlString = ""
+    @FocusState private var focused: Bool?
+    @State private var showInvalidURLError = false
+
+    let connectAction: (URL) -> Void
```
TimoPtr, line 10: Do it needs to be public?

Sources/App/Onboarding/Screens/OnboardingServersList/ManualURLEntry/ManualURLEntryView.swift
```diff
@@ -0,0 +1,102 @@
...
+                Alert(
+                    title: Text(verbatim: L10n.Onboarding.ManualSetup.InputError.title),
+                    message: Text(verbatim: L10n.Onboarding.ManualSetup.InputError.message),
+                    dismissButton: .default(Text(verbatim: L10n.okLabel))
+                )
+            }
+        }
+    }
+
+    @ViewBuilder
```
TimoPtr, line 50: I would add a comment on this function I had to run the app to understand what was it all about.

Sources/App/Onboarding/Screens/OnboardingServersList/ManualURLEntry/ManualURLEntryView.swift
```diff
@@ -0,0 +1,102 @@
...
+        }
+    }
+
+    private var connectButton: some View {
+        Button {
+            if let url = URL(string: urlString.trimmingCharacters(in: .whitespacesAndNewlines)) {
+                dismiss()
+                connectAction(url)
+            } else {
+                showInvalidURLError = true
```
TimoPtr, line 86: What will trigger that? I'm under the impression that URL is super permissive.

Sources/App/Onboarding/Screens/OnboardingServersList/OnboardingScanningInstanceRow.swift
```diff
@@ -0,0 +1,75 @@
...
+    var body: some View {
+        HStack {
+            icon
+            VStack(alignment: .leading) {
+                Text(name)
+                    .font(.headline)
+                Text(internalURLString ?? internalOrExternalURLString)
+                    .font(.subheadline)
+                    .foregroundColor(.secondary)
+                if let externalURLString {
```
TimoPtr, line 20: In your sample the URLs are the same, does it happens in real life? If so do you still want to display both?

Sources/App/Onboarding/Screens/OnboardingServersList/OnboardingScanningInstanceRow.swift
```diff
@@ -0,0 +1,75 @@
...
+
+    private var icon: some View {
+        ZStack {
+            Image(uiImage: Asset.SharedAssets.logo.image)
+                .resizable()
+                .aspectRatio(contentMode: .fit)
+                .frame(width: 30, height: 30)
+                .padding(.trailing, Spaces.one)
+            if internalURLString == nil, externalURLString != nil {
+                Image(systemSymbol: .icloudCircleFill)
```
TimoPtr, line 42: Do we want to use the iCloud icon here?

Sources/App/Onboarding/Screens/OnboardingServersList/OnboardingServersListView.swift
```diff
@@ -0,0 +1,156 @@
...
+            viewModel.startDiscovery()
+        }
+    }
+
+    private func onDisappear() {
+        viewModel.stopDiscovery()
+        viewModel.currentlyInstanceLoading = nil
+    }
+
+    @ViewBuilder
```
TimoPtr, line 67: Just curious why does you need some annotations sometimes?

Sources/App/Onboarding/Screens/OnboardingServersList/OnboardingServersListView.swift
```diff
@@ -0,0 +1,156 @@
...
+                server: nil,
+                error: error,
+                showSettingsEntry: false,
+                expandMoreDetails: true
+            )
+            .onDisappear {
+                viewModel.resetFlow()
+            }
+        } else {
+            Text("Unmapped onboarding flow (1)")
```
TimoPtr, line 80: no translation? I get that it should not happen but if it does it's going to be weird no?

Sources/App/Onboarding/Screens/OnboardingServersList/OnboardingServersListView.swift
```diff
@@ -0,0 +1,156 @@
+import Combine
+import Shared
+import SwiftUI
+
+struct OnboardingServersListView: View {
+    @Environment(\.dismiss) private var dismiss
+    @Environment(\.horizontalSizeClass) private var sizeClass
+
+    @EnvironmentObject var hostingProvider: ViewControllerProvider
```
TimoPtr, line 9: Do it needs to be public? It doesn't seem to be used outside of this struct

Sources/App/Onboarding/Screens/OnboardingServersList/OnboardingServersListView.swift
```diff
@@ -0,0 +1,156 @@
...
+        }
+        .navigationTitle(L10n.Onboarding.Scanning.title)
+        .navigationBarTitleDisplayMode(.inline)
+        .toolbar(content: {
+            ToolbarItem(placement: .topBarTrailing) {
+                ProgressView()
+                    .progressViewStyle(.circular)
+                    .opacity(viewModel.isLoading ? 1 : 0)
+                    .animation(.easeInOut, value: viewModel.isLoading)
+            }
```
TimoPtr, line 35: When do we see this?

Sources/App/Onboarding/Screens/OnboardingServersList/OnboardingServersListViewModel.swift
```diff
@@ -0,0 +1,125 @@
...
+    @Published var currentlyInstanceLoading: DiscoveredHomeAssistant?
+
+    @Published var showError = false
+    @Published var error: Error?
+
+    @Published var showPermissionsFlow = false
+    @Published var onboardingServer: Server?
+
+    /// Indicator for manual input loading
+    @Published var isLoading = false
```
TimoPtr, line 23: If it's specific for manual maybe the name should reflect it.

Sources/App/Resources/en.lproj/Localizable.strings
```diff
@@ -478,8 +484,9 @@ Tags will work on any device with Home Assistant installed which has hardware su
 "onboarding.welcome.description" = "This app connects to your Home Assistant server and allows integrating data about you and your phone.\
 \
 Home Assistant is free and open source home automation software with a focus on local control and privacy.";
-"onboarding.welcome.learn_more" = "Learn more";
```
TimoPtr, line 481: Will this gets removed from Localized once it's merged?

Sources/App/Resources/en.lproj/Localizable.strings
```diff
@@ -1179,4 +1186,13 @@ Home Assistant is free and open source home automation software with a focus on
 "widgets.sensors.description" = "Display state of sensors";
 "widgets.sensors.not_configured" = "No Sensors Configured";
 "widgets.sensors.title" = "Sensors";
-"yes_label" = "Yes";
\ No newline at end of file
+"yes_label" = "Yes";
+"onboarding.permission.location.description" = "To identify if you are at home and connect locally to Home Assistant, Apple requires that we ask for your location permission.";
+"onboarding.permission.location.privacy_note" = "Your location won't be shared with your local Home Assistant server unless you select 'Allow & Share with Home Assistant'. You can choose to share later in companion app settings > sensors.";
```
TimoPtr, line 1191: Probably not the scope of the PR but I think it's not super clear if I choose one option or the other. What do I loose if I select the second option VS the first.

Sources/App/WebView/WebViewController.swift
```diff
@@ -1358,3 +1360,53 @@ extension WebViewController {
         hud.hide(animated: true, afterDelay: 1.0)
     }
 }
+
```
TimoPtr, line 1363: I see that this file is also quite big like on Android ^^'

Sources/App/WebView/WebViewController.swift
```diff
@@ -128,6 +128,8 @@ final class WebViewController: UIViewController, WKNavigationDelegate, WKUIDeleg
         if #available(iOS 16.4, *) {
             webView.isInspectable = true
         }
+
+        postOnboardingNotificationPermission()
```
TimoPtr, line 132: Is this only going to be invoked after the onboarding?

Sources/App/WebView/WebViewController.swift
```diff
@@ -1358,3 +1360,53 @@ extension WebViewController {
         hud.hide(animated: true, afterDelay: 1.0)
     }
 }
+
+// MARK: - Post onboarding
+
+extension WebViewController {
+    private func postOnboardingNotificationPermission() {
+        DispatchQueue.main.asyncAfter(deadline: .now() + 3) { [weak self] in
```
TimoPtr: 3 looks like a magic number to me

Sources/Shared/API/Responses/DiscoveredHomeAssistant.swift
```diff
@@ -3,7 +3,7 @@ import ObjectMapper
 import PromiseKit
 import Version
 
-public struct DiscoveredHomeAssistant: ImmutableMappable {
+public struct DiscoveredHomeAssistant: ImmutableMappable, Equatable {
     public var uuid: String?
```
TimoPtr, line 7: Does it make sense to keep nullable?

Sources/Shared/API/Webhook/Sensors/AppVersionSensor.swift
```diff
@@ -10,7 +10,7 @@ final class AppVersionSensor: SensorProvider {
     func sensors() -> Promise<[WebhookSensor]> {
         let sensor = WebhookSensor(
             name: "App Version",
-            uniqueID: "app-version",
+            uniqueID: WebhookSensorId.appVersion.rawValue,
```
TimoPtr, line 13: This seems a bit out of the scope of the PR. But it seems to be minor changes.

Sources/Shared/Common/Extensions/View+HA.swift
```diff
@@ -0,0 +1,16 @@
+import Foundation
+import SwiftUI
+
+public extension View {
+    func embeddedInHostingController() -> UIHostingController<some View> {
```
TimoPtr, line 5: I don't know what this is and why we do that.

Sources/Shared/DesignSystem/Components/PrivacyNoteView.swift
```diff
@@ -0,0 +1,118 @@
+import SwiftUI
+
+public struct PrivacyNoteView: View {
```
TimoPtr, line 4: Do you want to limit this to Privacy? It could be a generic NoteView no?

Also I think you could add a general documentation of this component. For instance I didn't catch at first why there is a timer in it.

Sources/Shared/DesignSystem/Components/PrivacyNoteView.swift
```diff
@@ -0,0 +1,118 @@
...
+    @State private var background: AnyView
+    @State private var startPoint: UnitPoint = .topLeading
+    @State private var endPoint: UnitPoint = .bottomTrailing
+    @State private var timer: Timer?
+
+    let cornerRadius: CGFloat = 10
```
TimoPtr: Don't you have this somewhere defined globally?

Sources/Shared/DesignSystem/Components/PrivacyNoteView.swift
```diff
@@ -0,0 +1,118 @@
...
+            reverse()
+            startTimer()
+        }
+        .onDisappear {
+            stopTimer()
+        }
+    }
+
+    private func startTimer() {
+        timer = Timer.scheduledTimer(withTimeInterval: 2, repeats: true) { _ in
```
TimoPtr, line 62: To refer to my doc. You should specify the unit of the time here. either in a strong typing manner but I think there are no such thing in swift or by giving an explicit name or at least a comment?

Sources/Shared/DesignSystem/Styles/HAButtonStyles.swift
```diff
@@ -2,25 +2,88 @@ import Foundation
...
+
     public func makeBody(configuration: Configuration) -> some View {
         configuration.label
             .font(.callout.bold())
             .foregroundColor(.white)
             .frame(maxWidth: .infinity)
             .frame(height: 55)
-            .background(Color.asset(Asset.Colors.haPrimary))
+            .background(isEnabled ? Color.asset(Asset.Colors.haPrimary) : Color.gray)
+            .clipShape(RoundedRectangle(cornerRadius: 12))
```
TimoPtr: Radius is 12 here and 10 in privacy note (if I'm properly reading the code and understanding the API).

Sources/Shared/DesignSystem/Styles/HAButtonStyles.swift
```diff
@@ -2,25 +2,88 @@ import Foundation
...
 
 public struct HAButtonStyle: ButtonStyle {
+    @Environment(\.isEnabled) private var isEnabled: Bool
+
     public func makeBody(configuration: Configuration) -> some View {
         configuration.label
             .font(.callout.bold())
             .foregroundColor(.white)
             .frame(maxWidth: .infinity)
             .frame(height: 55)
```
TimoPtr: I would extract this height into a const since it's used multiple times.

Sources/Shared/DesignSystem/Styles/HAButtonStyles.swift
```diff
@@ -2,25 +2,88 @@ import Foundation
...
-            .background(Color.asset(Asset.Colors.haPrimary))
+            .background(isEnabled ? Color.asset(Asset.Colors.haPrimary) : Color.gray)
+            .clipShape(RoundedRectangle(cornerRadius: 12))
+            .opacity(isEnabled ? 1 : 0.5)
```
TimoPtr: I would extract the opacity value into a const.

Sources/Shared/DesignSystem/Styles/HAButtonStyles.swift
```diff
@@ -2,25 +2,88 @@ import Foundation
 import SwiftUI
 
 public struct HAButtonStyle: ButtonStyle {
```
TimoPtr, line 10: Do you have a showcase screen in the app that shows all the variants of the buttons? I would like to make one on Android when we are going to work on the Design System. Something accessible only in Debug from within the developer settings.

bgoncal: No, it can only be previewed im Xcode

Sources/Shared/DesignSystem/Styles/HAButtonStyles.swift
bgoncal: They are different elements, they don't require the same corner radius

Sources/Shared/DesignSystem/Components/PrivacyNoteView.swift
bgoncal: The scheduledTimer function asks me for a time interval and I pass it as 2, I think it is clear for iOS devs

Sources/App/Onboarding/Screens/OnboardingErrorView.swift
bgoncal: Currently nothing is displayed to the user, this feature is not being improved in this PR, only migrated in  to receive the onboarding in SwiftUI

Sources/Shared/DesignSystem/Styles/HAButtonStyles.swift
TimoPtr: You cannot create a dummy screen with all the kind of buttons?

Sources/App/Onboarding/Screens/OnboardingPermissions/LocationPermissionView.swift
bgoncal: When it is small enough its fine, but this one grew while I was implementing so I should move to a different file indeed.

Sources/App/Onboarding/Screens/OnboardingPermissions/OnboardingPermissionsNavigationView.swift
bgoncal: Its how we decide if the user is at home to access the local URL or not

Sources/App/Onboarding/Screens/OnboardingServersList/ManualURLEntry/ManualURLEntryView.swift
bgoncal: Nope, because this screen is what the name says a ManualURLEntryView, the other is not. The Navigation doest need to be in the name, I put there for the other file because that file is a container os possible screens of a selected group which will navigate given need.

Sources/App/Onboarding/Screens/OnboardingServersList/ManualURLEntry/ManualURLEntryView.swift
bgoncal: In a struct you can only set a non-mutable during initialization, so it's irrelevant to make it private

Sources/App/Onboarding/Screens/OnboardingServersList/ManualURLEntry/ManualURLEntryView.swift
bgoncal: If the input from the user is not a swift URL it will display an error.

Sources/App/Onboarding/Screens/OnboardingServersList/OnboardingScanningInstanceRow.swift
bgoncal: Well, that's why I added it there :D it's just a cloud icon provided by SFSymbols, so it can be used without worries. In this context it is just a representation of a server that has cloud connection

Sources/App/Onboarding/Screens/OnboardingServersList/OnboardingServersListView.swift
bgoncal: If you return just 1 view you dont need veiw builder, when you want to create views as if they were in the main body then you need it, like... one view below the other without putting them on a group.

Sources/App/Onboarding/Screens/OnboardingServersList/OnboardingServersListView.swift
bgoncal: For the user it is irrelevant since there are no instructions either way, a proper error screen will be done afterwards

Sources/App/Onboarding/Screens/OnboardingServersList/OnboardingServersListView.swift
bgoncal: EnvironmentObject can't be set from outside ona structure, so it is irrelevant

Sources/App/Onboarding/Screens/OnboardingServersList/OnboardingServersListView.swift
bgoncal: Good catch, it's not needed anymore

Sources/App/Onboarding/Screens/OnboardingServersList/OnboardingServersListViewModel.swift
bgoncal: Not needed anymore

Sources/App/Resources/en.lproj/Localizable.strings
bgoncal: Lokalise you mean? yes

Sources/App/Resources/en.lproj/Localizable.strings
bgoncal: The screen context makes it clear, tapping allow and share shares with home assistant, tappind allow for app only shares witht he app to execute what the subtitle says.

Sources/App/WebView/WebViewController.swift
bgoncal: Yes

Sources/Shared/API/Webhook/Sensors/AppVersionSensor.swift
bgoncal: It's just to have an enum of sensors IDs

Sources/Shared/API/Responses/DiscoveredHomeAssistant.swift
bgoncal: Good question, it was implemented already so I didn't touch assuming we cant guarantee the id

Sources/Shared/Common/Extensions/View+HA.swift
bgoncal: Well, it's iOS specific so thats why. It's to embed SwiftUI in UIKit code

Sources/Shared/DesignSystem/Components/PrivacyNoteView.swift
bgoncal: Depends, if I have elements which share the same radius yes, this one is a new component unrelated to the frontend created by me to highlight privacy, so it does not share the radius.

Sources/Shared/DesignSystem/Styles/HAButtonStyles.swift
bgoncal: I can of course, but I won't in this PR, maybe in the future when we have a proper Design system for mobile.

bgoncal:
> Quite a big PR 👏🏻 I don't see any test is that on purpose? Do we have screenshot testings on this new screens?

Still coming in this PR

Sources/App/Onboarding/Container/OnboardingNavigationView.swift
bgoncal: Sorry I missed your second comment and replied over it, yes for now location only, but this flow is not final, I removed motion and focus sensor, they may return before the release, this will stay flexible in case we have more permissions later.

Sources/App/Onboarding/Screens/OnboardingServersList/ManualURLEntry/ManualURLEntryView.swift
bgoncal: 192.168.0.1 without http would trigger the error

Sources/App/Onboarding/Screens/OnboardingServersList/OnboardingScanningInstanceRow.swift
bgoncal: I've seen it happening in Android only, in iOS I never seen external URL available for some reason, I added support only because I saw it in Android. For now I'll display both URLs and not add any extra logic.

Sources/App/WebView/WebViewController.swift
bgoncal: It was even bigger, I already removed a bunch from it 😂

bgoncal: There is a migration for this screen planned as well, but it is a big task so it will need to be splitted and will take long time

Sources/App/WebView/WebViewController.swift
bgoncal: And it is :) Feels a good timin to display it after onboarding, no need to be this number, but more of less than this feels weird when using the app

Sources/Shared/DesignSystem/Components/PrivacyNoteView.swift
bgoncal: No I don't want a note component for now, for now I want a privacy note component with the criteria I added to this component, if needed I'll create a separate component.

The timer is there for the animation, I won't add comments for all parts of the code, some interpretation needs to be done while using it.

bgoncal: About documentation, after we have a proper design system a documentation makes sense, before that it doesn't since components can (and will) change anytime.

Sources/Shared/DesignSystem/Styles/HAButtonStyles.swift
TimoPtr: Sure I forgot to mention that it was not a request.

Sources/App/WebView/WebViewController.swift
TimoPtr: You should probably add a comment then about it. Saying that number seems to be the right number so it looks nice on the app. I think it's a valuable information for a developer coming by.

Sources/Shared/DesignSystem/Components/PrivacyNoteView.swift
TimoPtr: I think documentation in this case should be about the generate behavior and what is this component for. If we stay away from the impl details it should help keeps it up to date even if the internals changes.

It's more a general comments. But fine by me if you prefer to keep like this.

Sources/App/WebView/WebViewController.swift
bgoncal: I'll add a comment

Sources/Shared/DesignSystem/Components/PrivacyNoteView.swift
bgoncal: I'll add a header comment to it then.

Sources/App/Onboarding/Screens/OnboardingServersList/OnboardingServersListView.swift
bgoncal: Apologies, it is needed for manual URL entry since it doesn't have a row in the list to display the loader, then it displays it in the top navigation bar.

PR #3784, opened 2025-08-31
Fixup misleading nomenclature: s/frontend cache/client cache/g

nethershaw:
## Summary

This commit replaces every instance of a misleading reference to a "frontend cache" where there is no such thing with a correct, however generalized, reference to the "client cache."

In objective fact, a "frontend cache" is a term reserved for server-side caching reverse proxies centralized on and attributable to a frontend itself. No such resource exists in any Home Assistant deployment. This incorrect use of terms is confusing and misleading, and therefore is herein replaced. I will confess to having been thusly misled despite (and perhaps _because of_) being employed in the relevant technology industries for going on two decades.

The only relevant cache is attributable to the client, whether it is a browser or whether it is a dedicated platform client. This change categorically prevents the mistaken apprehension that any server-side cache attributable to the frontend requires resetting, and at the very least, is *less incorrect.* If the word "client" is objectionable, valid alternatives e.g. "browser" and "user," or simply eschewing the misleading qualifier entirely (not recommended), remain available.

This change makes absolutely no attempt to alter localised strings where the word "frontend" has already been translated. Wheresoever these strings exist, it remains the responsibility of the repository maintainer to find and reconcile them. You may choose whatever replacements you like; but these must not be misleading to the point of falsely indicating the presence of an application component.

There remain functions whose names reference a FrontendCache; while these names remain properly incorrect, this PR also makes absolutely no attempt to rename them, since they (a) do not affect the user and (b) I am not interested in refactoring anyone's actual code and maintaining referential integrity, which would likely require further updates to unit tests, and so on. Honestly, the only such functions I have identified are `resetFrontendCacheIfNeeded()` and `serverNeedsFrontendReset()` (which is even worse; you aren't resetting the frontend here at all!) in WebViewController.swift, but I am not going to touch them. If consistency requires correction of even this mistake, edits and expansions to this pull request will be gratefully accepted. I strongly suggest, however, you take the user-facing linguistic win as an atomic operation.

## Screenshots

It is a misuse of a single word. Screenshots of every appearance will not be provided.

No string matching the case-insensitive expression "frontend cache" appears in the documentation (indeed, this is the reason for the other corrections); therefore, no correction is required here.

## Any other notes

cf. https://community.home-assistant.io/t/how-to-clear-the-frontend-cache/670491/91

This pull request is part of a set against the following repositories (crosslinks may be updated after the PR is opened):
  - [home-assistant/android](https://github.com/home-assistant/android/pull/5711)
  - [home-assistant/iOS](https://github.com/home-assistant/iOS/pull/3784) (you are here)
  - [hacs/integration](https://github.com/hacs/integration/pull/4788)

### DIFF
--- a/Sources/App/Resources/bg.lproj/Localizable.strings
+++ b/Sources/App/Resources/bg.lproj/Localizable.strings
@@ -663,7 +663,7 @@ Home Assistant е безплатен и с отворен код софтуер 
 "settings.reset_section.reset_alert.title" = "Нулиране";
 "settings.reset_section.reset_app.title" = "Reset App (Remove servers and data)";
 "settings.reset_section.reset_row.title" = "Нулиране";
-"settings.reset_section.reset_web_cache.title" = "Reset frontend cache";
+"settings.reset_section.reset_web_cache.title" = "Reset client cache";
 "settings.server_select.page_title" = "Сървър";
 "settings.server_select.title" = "Сървър";
 "settings.status_section.header" = "Състояние";
--- a/Sources/App/Resources/cy-GB.lproj/Localizable.strings
+++ b/Sources/App/Resources/cy-GB.lproj/Localizable.strings
@@ -663,7 +663,7 @@ Home Assistant is free and open source home automation software with a focus on 
 "settings.reset_section.reset_alert.title" = "Ailosod";
 "settings.reset_section.reset_app.title" = "Reset App (Remove servers and data)";
 "settings.reset_section.reset_row.title" = "Ailosod";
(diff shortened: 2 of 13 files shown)


nethershaw:
/cla-recheck

okay, or don't

bgoncal:
Waiting for a conclusion from the discussion happening in the Android repo.

bgoncal:
Please reopen the PR when a decision has been made on this regard, for now I consider frontend cache good enough for the App.

When reopening this PR make sure no localizable string is changed directly and instead use Lokalise for that.

PR #4434, opened 2026-03-13
Fix kiosk screensaver idle timer not resetting on touch

nstefanelli:
## Summary
- Kiosk screensaver idle timer was not resetting on user interaction — `recordKioskActivity()` was defined but never called
- WKWebView consumes touch events before UIKit idle detection, so the timer counted down from kiosk activation rather than last touch
- Added a tap gesture recognizer on the WebView (`cancelsTouchesInView = false`) to detect taps without interfering with normal interaction
- Added `scrollViewWillBeginDragging` delegate hook to detect scroll/drag interactions

## Test plan
- [ ] Enable kiosk mode with 30s screensaver timeout
- [ ] Tap on the WebView dashboard — idle timer should reset
- [ ] Scroll on the dashboard — idle timer should reset
- [ ] Stop interacting — screensaver should appear after configured timeout
- [ ] Verify normal WebView interaction (taps, links, scrolling) still works normally

### DIFF
--- a/Sources/App/Frontend/WebView/WebViewController.swift
+++ b/Sources/App/Frontend/WebView/WebViewController.swift
@@ -237,6 +237,7 @@ final class WebViewController: UIViewController, WKNavigationDelegate, WKUIDeleg
 
         webView.navigationDelegate = self
         webView.uiDelegate = self
+        webView.scrollView.delegate = self
 
         setupWebViewConstraints(statusBarView: statusBarView)
         setupPullToRefresh()


nstefanelli:
@bgoncal think you could do a quick review to get this merged in for TestFlight testing?

nstefanelli:
**Note on `webView.scrollView.delegate = self`:** This also enables the pre-existing `scrollViewDidScroll` (scroll bounds checking) in `WebViewController+WebKitDelegates.swift` which was previously unreachable without the delegate assignment. Both the existing scroll bounds logic and the new `scrollViewWillBeginDragging` hook now work correctly.

PR #4442, opened 2026-03-19
Fix using elementId in handleElementFocus script

wendevlin:
## Summary
- Fix home-assistant/frontend#30048
- elementId inside the JS was used in the return string, but not escaped so JS fails.
  -  I am actually not sure if this return string is even needed. Does it have any sense?

## Any other notes
- when this is releases we activate again the functionality in the frontend: home-assistant/frontend#30226


bgoncal:
@wendevlin It's not really needed I guess, it was more a debug thing, I will remove it

PR #4444, opened 2026-03-19
Adds iOS Live Activities support

rwarner:
## Summary

> For architecture decisions, data model details, iOS version strategy, push token flow, and rate limiting — see 
[technical-brief.pdf](https://github.com/user-attachments/files/26280274/technical-brief.pdf)

Adds iOS Live Activities support, letting Home Assistant automations push real-time state to the Lock Screen — washing machine countdowns, EV charging progress, delivery tracking, alarm states, or anything time-sensitive that benefits from glanceable visibility without unlocking the phone.

When an automation sends a notification with `live_update: true` in the data payload, the companion app starts a Live Activity instead of (or in addition to) a standard notification banner. Subsequent pushes with the same `tag` update it in-place silently. `clear_notification` + `tag` ends it.

Field names (`tag`, `title`, `message`, `progress`, `progress_max`, `chronometer`, `when`, `when_relative`, `notification_icon`, `notification_icon_color`) are intentionally shared with Android's Live Notifications API. Both platforms use the same `live_update: true` trigger — a single YAML block targets iOS 17.2+ and Android 16+ with no platform-specific keys.

```yaml
data:
  title: "Washing Machine"
  message: "Cycle in progress"
  data:
    tag: washer_cycle
    live_update: true           # Android 16+ and iOS 17.2+
    progress: 2700
    progress_max: 3600
    chronometer: true
    when: 2700
    when_relative: true
    notification_icon: mdi:washing-machine
    notification_icon_color: "#2196F3"
```

**New files:**
- `Sources/Shared/LiveActivity/HALiveActivityAttributes.swift` — the `ActivityAttributes` type. Field names match the Android payload spec. **Struct name and `CodingKeys` are wire-format frozen** — APNs push-to-start payloads reference the Swift type name directly.
- `Sources/Shared/LiveActivity/LiveActivityRegistry.swift` — Swift `actor` managing `Activity<HALiveActivityAttributes>` lifecycle. Uses a reservation pattern to prevent duplicate activities when two pushes with the same `tag` arrive simultaneously.
- `Sources/Shared/Notifications/NotificationCommands/HandlerLiveActivity.swift` — start/update and end `NotificationCommandHandler` implementations, guarded against the `PushProvider` extension process where ActivityKit is unavailable.
- `Sources/Extensions/Widgets/LiveActivity/` — `ActivityConfiguration` wrapper, Lock Screen / StandBy view, and compact / minimal / expanded Dynamic Island views.
- `Sources/App/Settings/LiveActivity/LiveActivitySettingsView.swift` — activity status, active list, privacy disclosure, and 11 debug scenarios for pre-server-side testing.

**Modified files:**
- `Widgets.swift` — registers `HALiveActivityConfiguration` in all three `WidgetBundle` variants behind `#available(iOSApplicationExtension 17.2, *)`
- `NotificationsCommandManager.swift` — registers new handlers; `HandlerClearNotification` now also ends a matching Live Activity when `tag` is present
- `HAAPI.swift` — adds `supports_live_activities`, `supports_live_activities_frequent_updates`, `live_activity_push_to_start_token`, `live_activity_push_to_start_apns_environment` to registration payload under a single `#available(iOS 17.2, *)` check
- `AppDelegate.swift` — reattaches surviving activities at launch and starts observing push-to-start tokens under a single `#available(iOS 17.2, *)` check
- `Info.plist` — `NSSupportsLiveActivities` + `NSSupportsLiveActivitiesFrequentUpdates`
- `SettingsItem.swift` / `SettingsView.swift` — Live Activities settings row is gated behind `Current.isTestFlight` and shows a `BetaLabel` badge

**Tests:** 
 - *Unit Tests*
   - 45 new tests (36 handler tests, 9 command manager routing tests). All 490 existing tests continue to pass.
 - *Device Tests*
   - I tried to test with my physical device, however I do not have a paid account. Turns out I could not deploy without disabling a lot of entitlements for compilation. Even so, once I did get it deployed and running Live Activities wouldn't show unless some of those settings that I disable were re-enable and leaving me in a particular spot.
   - I mainly tested with the simulator which is what is shown below

## Screenshots

full-debug-scenarios

Documentation: home-assistant/companion.home-assistant#1303

## Link to pull request in push relay repository

Relay server: home-assistant/mobile-apps-fcm-push#278

## Link to pull request in HA core

Core: home-assistant/core#166072

## Any other notes

**iOS version gating at 17.2.** The entire feature is gated at `#available(iOS 17.2, *)` — this is the minimum for push-to-start (the primary server-side start mechanism). A single availability check now covers all Live Activity APIs: `supports_live_activities`, `frequentPushesEnabled`, push-to-start token, and all ActivityKit usage. This eliminates the nested 16.2/17.2 check pattern.

**Push-to-start (iOS 17.2+) is client-complete.** The token is observed, stored in Keychain, and included in registration payloads. All companion server-side PRs are now open — relay server at home-assistant/mobile-apps-fcm-push#278 and HA core webhook handlers at home-assistant/core#166072. The relay server uses FCM's native `apns.liveActivityToken` support (Firebase Admin SDK v13.5.0+) — no custom APNs client or credentials needed.

> **Server-side work** — all PRs now open:
> - ~~`supports_live_activities` field handling in device registration~~ ✓ home-assistant/core#166072
> - ~~`mobile_app_live_activity_token` webhook handler~~ ✓ home-assistant/core#166072
> - ~~`mobile_app_live_activity_dismissed` webhook handler~~ ✓ home-assistant/core#166072
> - ~~Relay server: Live Activity delivery via FCM `apns.liveActivityToken`~~ ✓ home-assistant/mobile-apps-fcm-push#278
> - ~~`notify.py` routing: includes Live Activity APNs token alongside FCM token~~ ✓ home-assistant/core#166072

**Live Activities entry in Settings is gated behind TestFlight.** The settings row only appears when `Current.isTestFlight` is true, preventing it from surfacing in a release build before the feature is fully tested. A `BetaLabel` badge is shown alongside the row title.

**iPad:** `areActivitiesEnabled` is always `false` on iPad — Apple system restriction. The Settings screen shows "Not available on iPad." The registry silently no-ops. HA receives `supports_live_activities: false` in the device registration for iPad.

**`HALiveActivityAttributes` is frozen post-ship.** The struct name appears as `attributes-type` in APNs push-to-start payloads. Renaming it silently breaks all remote starts. The `ContentState` `CodingKeys` are equally frozen — only additions are safe. Both have comments in the source calling this out.

**The debug section in Settings is intentional.** Gated behind `#if DEBUG` so it only appears in debug builds — it never ships to TestFlight or the App Store. It exercises the full ActivityKit lifecycle without requiring the server-side chain.

**`UNUserNotificationCenter` in tests.** The `clear_notification` + `tag` → Live Activity dismissal path is covered by code review rather than a unit test. `HandlerClearNotification` calls `UNUserNotificationCenter.current().removeDeliveredNotifications` synchronously, which requires a real app bundle and throws `NSInternalInconsistencyException` in the XCTest host. A comment in the test file explains this.

**Rate limiting on iOS 18.** Apple throttles Live Activity updates to ~15 seconds between renders. Automations should trigger on state change events, not polling timers.

**Related:**
- Community discussion: https://github.com/orgs/home-assistant/discussions/84
- Android companion reference: https://github.com/home-assistant/android
- Roadmap: https://github.com/home-assistant/roadmap/issues/52

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -67,7 +67,7 @@ jobs:
 
       - name: Comment on PR if `swiftlint:disable` is found
         if: steps.scan.outputs.disable_lines != '' && github.event.pull_request.head.repo.full_name == github.repository
-        uses: marocchino/sticky-pull-request-comment@70d2764d1a7d5d9560b100cbea0077fc8f633987  # v3.0.2
+        uses: marocchino/sticky-pull-request-comment@d4d6b0936434b21bc8345ad45a440c5f7d2c40ff  # v3.0.3
         with:
           header: swiftlint-disable-check
           message: |
@@ -112,7 +112,7 @@ jobs:
         if: |
           steps.detect.outputs.has_unused == 'true' &&
           github.event.pull_request.head.repo.full_name == github.repository
-        uses: marocchino/sticky-pull-request-comment@70d2764d1a7d5d9560b100cbea0077fc8f633987  # v3.0.2
+        uses: marocchino/sticky-pull-request-comment@d4d6b0936434b21bc8345ad45a440c5f7d2c40ff  # v3.0.3
         with:
           header: unused-strings-check
           message: |
--- a/.github/workflows/push_deploy.yml
+++ b/.github/workflows/push_deploy.yml
@@ -20,7 +20,7 @@ jobs:
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd  # v6.0.2
 
       - name: Login to GitHub Container Registry
-        uses: docker/login-action@b45d80f862d83dbcd57f89517bcf500b2ab88fb2  # v4.0.0
+        uses: docker/login-action@4907a6ddec9925e35a0a9e82d7399ccc52663121  # v4.1.0
         with:
           registry: ghcr.io
           username: ${{ github.repository_owner }}
--- a/HomeAssistant.xcodeproj/project.pbxproj
+++ b/HomeAssistant.xcodeproj/project.pbxproj
@@ -1230,6 +1230,7 @@
 		46F103262D721516002BC586 /* LocationHistoryListViewSnapshot.test.swift in Sources */ = {isa = PBXBuildFile; fileRef = 46F103252D721504002BC586 /* LocationHistoryListViewSnapshot.test.swift */; };
 		491E98FF25D543560077BBE3 /* LogbookEntry.swift in Sources */ = {isa = PBXBuildFile; fileRef = 491E98FE25D543560077BBE3 /* LogbookEntry.swift */; };
 		491E990025D543560077BBE3 /* LogbookEntry.swift in Sources */ = {isa = PBXBuildFile; fileRef = 491E98FE25D543560077BBE3 /* LogbookEntry.swift */; };
+		504A2C862F8133E6002A3C0E /* CarPlayTabsSelectionView.swift in Sources */ = {isa = PBXBuildFile; fileRef = 504A2C852F8133E6002A3C0E /* CarPlayTabsSelectionView.swift */; };
 		51B609B13F56C4F17CEF2BE4 /* KioskConstants.swift in Sources */ = {isa = PBXBuildFile; fileRef = 825E1E44BA9ABF1BF53733D3 /* KioskConstants.swift */; };
 		54B2C38995814098B677135A /* WidgetCommonlyUsedEntities.swift in Sources */ = {isa = PBXBuildFile; fileRef = DEEEA3344F064DB183F46C47 /* WidgetCommonlyUsedEntities.swift */; };
 		58213D5B1792311CD4CA261D /* Pods_iOS_Extensions_NotificationService.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = F6DA82FEEE2DDC3B2CC20DA3 /* Pods_iOS_Extensions_NotificationService.framework */; };
@@ -3046,6 +3047,7 @@
 		49BCC46A2F929005F9529BF9 /* Pods-watchOS-WatchExtension-Watch.beta.xcconfig */ = {isa = PBXFileReference; includeInIndex = 1; lastKnownFileType = text.xcconfig; name = "Pods-watchOS-WatchExtension-Watch.beta.xcconfig"; path = "Pods/Target Support Files/Pods-watchOS-WatchExtension-Watch/Pods-watchOS-WatchExtension-Watch.beta.xcconfig"; sourceTree = "<group>"; };
 		4A74A2EF6714F9B13C1DAFBD /* Pods-Tests-App.debug.xcconfig */ = {isa = PBXFileReference; includeInIndex = 1; lastKnownFileType = text.xcconfig; name = "Pods-Tests-App.debug.xcconfig"; path = "Pods/Target Support Files/Pods-Tests-App/Pods-Tests-App.debug.xcconfig"; sourceTree = "<group>"; };
 		4F28982D819D8E105E6FB878 /* Pods-Tests-App.release.xcconfig */ = {isa = PBXFileReference; includeInIndex = 1; lastKnownFileType = text.xcconfig; name = "Pods-Tests-App.release.xcconfig"; path = "Pods/Target Support Files/Pods-Tests-App/Pods-Tests-App.release.xcconfig"; sourceTree = "<group>"; };
+		504A2C852F8133E6002A3C0E /* CarPlayTabsSelectionView.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = CarPlayTabsSelectionView.swift; sourceTree = "<group>"; };
 		50D9C22ED2834EC9DAAC63AC /* Pods-iOS-Extensions-Intents.debug.xcconfig */ = {isa = PBXFileReference; includeInIndex = 1; lastKnownFileType = text.xcconfig; name = "Pods-iOS-Extensions-Intents.debug.xcconfig"; path = "Pods/Target Support Files/Pods-iOS-Extensions-Intents/Pods-iOS-Extensions-Intents.debug.xcconfig"; sourceTree = "<group>"; };
 		553A33E097387AA44265DB13 /* Pods-iOS-App-metadata.plist */ = {isa = PBXFileReference; includeInIndex = 1; lastKnownFileType = text.plist.xml; name = "Pods-iOS-App-metadata.plist"; path = "Pods/Pods-iOS-App-metadata.plist"; sourceTree = "<group>"; };
 		592EED7A6C2444872F11C17B /* Pods-iOS-Extensions-NotificationService-metadata.plist */ = {isa = PBXFileReference; includeInIndex = 1; lastKnownFileType = text.plist.xml; name = "Pods-iOS-Extensions-NotificationService-metadata.plist"; path = "Pods/Pods-iOS-Extensions-NotificationService-metadata.plist"; sourceTree = "<group>"; };
@@ -6268,6 +6270,7 @@
 			isa = PBXGroup;
 			children = (
 				42EFFAEB2C8882DD002F10FC /* CarPlayConfigurationView.swift */,
+				504A2C852F8133E6002A3C0E /* CarPlayTabsSelectionView.swift */,
 				42ABB0BA2C888BB10081461D /* CarPlayConfigurationViewModel.swift */,
 				42ABB0B82C888AA10081461D /* CarPlayConfig.swift */,
 			);
@@ -9457,6 +9460,7 @@
 				42462E692D9ED75900ECC8A7 /* LocationPermissionView.swift in Sources */,
 				42D0AE452D88259000D9715A /* WidgetDocumentationLink.swift in Sources */,
 				1100D51D2496AECE00B1073C /* PermissionStatusRow.swift in Sources */,
+				504A2C862F8133E6002A3C0E /* CarPlayTabsSelectionView.swift in Sources */,
 				42E6C08A2CE4F4FA007CA622 /* DownloadManagerView.swift in Sources */,
 				42E95C572CA45EFA0010ECE3 /* OnboardingErrorView.swift in Sources */,
 				429858882EB13FD300E33710 /* WebViewControllerOverlayedViewTags.swift in Sources */,
--- a/Sources/App/Settings/CarPlay/CarPlayConfigurationView.swift
+++ b/Sources/App/Settings/CarPlay/CarPlayConfigurationView.swift
@@ -160,60 +160,10 @@ struct CarPlayConfigurationView: View {
     private var tabsSection: some View {
         Section(L10n.CarPlay.Config.Tabs.title) {
             NavigationLink {
-                tabsSelection
+                CarPlayTabsSelectionView(viewModel: viewModel)
             } label: {
                 Text(viewModel.config.tabs.compactMap(\.name).joined(separator: ", "))
                     .frame(maxWidth: .infinity, alignment: .leading)
-            }
-        }
-    }
-
-    private var tabsSelection: some View {
-        List {
-            Section {
-                ForEach(viewModel.config.tabs, id: \.rawValue) { tab in
-                    Button {
-                        viewModel.updateTab(tab, active: false)
-                    } label: {
-                        HStack {
-                            Text(tab.name)
-                                .frame(maxWidth: .infinity, alignment: .leading)
-                            Image(systemSymbol: .line3Horizontal)
-                                .foregroundStyle(.gray)
-                        }
-                    }
-                    .buttonStyle(.plain)
-                }
-                .onMove { indices, newOffset in
-                    viewModel.moveTab(from: indices, to: newOffset)
-                }
-                .onDelete { indexSet in
-                    viewModel.deleteTab(at: indexSet)
-                }
-            } header: {
-                Text(L10n.CarPlay.Tabs.Active.title)
-            } footer: {
-                Text(L10n.CarPlay.Tabs.Active.DeleteAction.title)
-            }
-            if viewModel.config.tabs.count != CarPlayTab.allCases.count {
-                Section(L10n.CarPlay.Tabs.Inactive.title) {
-                    ForEach(CarPlayTab.allCases.filter({ tab in
-                        !viewModel.config.tabs.contains(tab)
-                    }), id: \.rawValue) { tab in
-                        Button {
-                            viewModel.updateTab(tab, active: true)
-                        } label: {
-                            HStack {
-                                Text(tab.name)
-                                    .frame(maxWidth: .infinity, alignment: .leading)
-                                Image(systemSymbol: .plusCircleFill)
-                                    .foregroundStyle(.white, .green)
-                                    .font(.title3)
-                            }
-                        }
-                        .buttonStyle(.plain)
-                    }
-                }
             }
         }
         .animation(.bouncy, value: viewModel.config.tabs)
--- a/Sources/App/Settings/CarPlay/CarPlayTabsSelectionView.swift
+++ b/Sources/App/Settings/CarPlay/CarPlayTabsSelectionView.swift
@@ -1,0 +1,61 @@
+import SFSafeSymbols
+import Shared
+import SwiftUI
+
+struct CarPlayTabsSelectionView: View {
+    @ObservedObject var viewModel: CarPlayConfigurationViewModel
+
+    var body: some View {
+        List {
+            Section {
+                ForEach(viewModel.config.tabs, id: \.rawValue) { tab in
+                    Button {
+                        viewModel.updateTab(tab, active: false)
+                    } label: {
+                        HStack {
+                            Text(tab.name)
+                                .frame(maxWidth: .infinity, alignment: .leading)
+                            Image(systemSymbol: .line3Horizontal)
+                                .foregroundStyle(.gray)
+                        }
+                    }
+                    .buttonStyle(.plain)
+                }
+                .onMove { indices, newOffset in
+                    viewModel.moveTab(from: indices, to: newOffset)
+                }
+                .onDelete { indexSet in
+                    viewModel.deleteTab(at: indexSet)
+                }
+            } header: {
+                Text(L10n.CarPlay.Tabs.Active.title)
+            } footer: {
+                Text(L10n.CarPlay.Tabs.Active.DeleteAction.title)
+            }
+            if viewModel.config.tabs.count != CarPlayTab.allCases.count {
+                Section(L10n.CarPlay.Tabs.Inactive.title) {
+                    ForEach(CarPlayTab.allCases.filter({ tab in
+                        !viewModel.config.tabs.contains(tab)
+                    }), id: \.rawValue) { tab in
+                        Button {
+                            viewModel.updateTab(tab, active: true)
+                        } label: {
+                            HStack {
+                                Text(tab.name)
+                                    .frame(maxWidth: .infinity, alignment: .leading)
+                                Image(systemSymbol: .plusCircleFill)
+                                    .foregroundStyle(.white, .green)
+                                    .font(.title3)
+                            }
+                        }
+                        .buttonStyle(.plain)
+                    }
+                }
+            }
+        }
+    }
+}
+
+#Preview {
+    CarPlayTabsSelectionView(viewModel: CarPlayConfigurationViewModel())
+}


Sources/Shared/Notifications/NotificationCommands/NotificationsCommandManager.swift
```diff
@@ -89,6 +108,23 @@ private struct HandlerClearNotification: NotificationCommandHandler {
...
+        #if os(iOS) && canImport(ActivityKit)
+        if #available(iOS 16.2, *), !Current.isAppExtension, let tag = payload["tag"] as? String {
+            return Promise<Void> { seal in
+                Task {
+                    await Current.liveActivityRegistry.end(tag: tag, dismissalPolicy: .immediate)
+                    // https://stackoverflow.com/a/56657888/6324550
+                    DispatchQueue.main.async { seal.fulfill(()) }
+                }
+            }
+        }
```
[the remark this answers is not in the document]
rwarner: The omission is documented in the PR description under "Any other notes." `UNUserNotificationCenter.current().removeDeliveredNotifications` throws             
  `NSInternalInconsistencyException` in the XCTest host — this is an Apple limitation, not specific to this change. Wrapping it in a protocol for testability is
  reasonable future work but out of scope here. The Live Activity dismissal path (`liveActivityRegistry.end`) is already covered by the existing registry tests.

bgoncal:
tip for lint you can use `bundle exec fastlane autocorrect`

bgoncal:
Can you make a screen recording showing the whole flow? Showing HA UI sending the push and iPhone handling it

rwarner:
> Can you make a screen recording showing the whole flow? Showing HA UI sending the push and iPhone handling it

So I don't have a paid Apple dev account so apparently the only way I could get it to compile to an actual device (to try to get push notifs) was to rip a bunch of entitlements out to compile for my physical device. For the emulator, I wasn't sure how to do a full-end-to-end test with the four repo changes altogether and run that locally. Theoretically it seemed like a possibility to do a local push to the emulated device but didn't get to that point.

Mainly why I created the debug options in the screenshot section showcasing all of the possibilities with the YAML at the moment. (This is in App Companion Settings -> Live Activity -> DEBUG options)

I don't mind putting some more stuff together, lemme know any specifics you would like to see and I can try. There's also more screenshots in the documentation repo here: https://github.com/home-assistant/companion.home-assistant/pull/1303

bgoncal:
Nowadays you can already test push notification on simulator as well from what I can remember, check this https://www.tiagohenriques.dev/blog/testing-push-notifications-ios-simulators

rwarner:
Thanks, I'll look at this and see if I can produce some more examples for you / video on top of the previous gif

rwarner:
> Nowadays you can already test push notification on simulator as well from what I can remember, check this https://www.tiagohenriques.dev/blog/testing-push-notifications-ios-simulators

~Thanks for the link! I looked into xcrun simctl push — from my testing, it delivers through UNUserNotificationCenter rather than ActivityKit's push token channel, so Live Activity payloads (event, content-state, etc.) arrive as regular notifications instead of updating the running activity. The simulator also returns nil for ActivityKit push tokens, so real APNs delivery to a Live Activity isn't possible either. Happy to be corrected if there's a way I'm missing!~

For the local WebSocket path, the simulator is currently connecting through Nabu Casa (cloud relay), so notifications route through FCM rather than the local WebSocket channel. I'd need to be on the same local network as the HA instance for WebSocket-based delivery to work — ~I'll try to set that up.~
**Update: Would need a new Hass instance running my new core feature branch**

In the meantime, I've added two screen recordings demonstrating start/update/end with various payload configurations on the simulator. The debug section in Settings also exercises all the UI states. As well as correcting the linting issues

https://github.com/user-attachments/assets/beb389d6-12e2-4a5c-86d1-eb95462fe7e8


https://github.com/user-attachments/assets/71b4bd7f-390c-453a-8cbd-b623bb864791

rwarner:
Progress with `xcrun simctl` actually, I was able to get a "push" to work using the following file and a small recent code change I just pushed

`live_activity_test.apns`
```
{
  "Simulator Target Bundle": "com.rwarner.ha.HomeAssistant.dev",
  "aps": {
    "alert": {
      "title": "Washing Machine",
      "body": "Cycle in progress"
    },
    "mutable-content": 1,
    "content-available": 1
  },
  "homeassistant": {
    "live_activity": true,
    "tag": "washer_cycle",
    "title": "Washing Machine",
    "message": "Cycle in progress",
    "progress": 2700,
    "progress_max": 3600,
    "chronometer": true,
    "when": 2700,
    "when_relative": true,
    "notification_icon": "mdi:washing-machine",
    "notification_icon_color": "#2196F3"
  }
}
```

`xcrun simctl push booted com.rwarner.ha.HomeAssistant.dev /tmp/live_activity_test.apns`

Which works while the application is in the foreground of the iPhone:

For some reason the dynamic island in the emulator is fussy but pops up very quickly
Screenshot 2026-03-23 at 3 38 23 PM

https://github.com/user-attachments/assets/32aed314-111f-4f04-9879-89e6cf90bd42

Tests: `push payload →
  NotificationCommandManager → HandlerStartOrUpdateLiveActivity → LiveActivityRegistry → ActivityKit → Lock Screen`
  
 However to test an **actual background push** Live Activity handler from what I can figure out I would need:
                                                                    
  - Local WebSocket — HA instance compiled with my `core` changes on same network                                                                                         
  - Real APNs — paid developer account                                                                                                                
  - FCM — production Firebase project matching the bundle ID
  - ?Apple's Push Notification console in their paid dev area?

bgoncal:
> Local WebSocket — HA instance compiled with my core changes on same network

You have that already right? Because you opened a PR for core as well, so you can use that instance you used to test core to connect to the iOS App

Sources/Extensions/Widgets/LiveActivity/HALockScreenView.swift
```diff
@@ -0,0 +1,77 @@
...
+/// Keep layout tight and avoid decorative spacing.
+@available(iOS 16.2, *)
+struct HALockScreenView: View {
+    let attributes: HALiveActivityAttributes
+    let state: HALiveActivityAttributes.ContentState
+
+    var body: some View {
+        VStack(alignment: .leading, spacing: 6) {
+            // Header row: icon + title
+            HStack(spacing: 8) {
```
bgoncal: Use DesignSystem.Spaces

Sources/Extensions/Widgets/LiveActivity/HALockScreenView.swift
```diff
@@ -0,0 +1,77 @@
...
+                    .font(.headline)
+                    .foregroundStyle(.white)
+                    .lineLimit(1)
+            }
+
+            // Body: timer or message
+            if state.chronometer == true, let end = state.countdownEnd {
+                Text(timerInterval: Date.now ... end, countsDown: true)
+                    .font(.subheadline)
+                    .foregroundStyle(.white.opacity(0.85))
```
bgoncal: Create a color asset to represent this color, just so we dont duplciate in the future if needed and keep it standard

Sources/Extensions/Widgets/LiveActivity/HALockScreenView.swift
```diff
@@ -0,0 +1,77 @@
...
+                    .monospacedDigit()
+            } else {
+                Text(state.message)
+                    .font(.subheadline)
+                    .foregroundStyle(.white.opacity(0.85))
```
bgoncal: Same here

Sources/Extensions/Widgets/LiveActivity/HALockScreenView.swift
```diff
@@ -0,0 +1,77 @@
...
+                    .lineLimit(2)
+            }
+
+            // Progress bar (only when progress data is present)
+            if let fraction = state.progressFraction {
+                ProgressView(value: fraction)
+                    .tint(accentColor)
+            }
+        }
+        .padding(.horizontal, 16)
```
bgoncal: Use DesignSystem.Spaces

Sources/Extensions/Widgets/LiveActivity/HALockScreenView.swift
```diff
@@ -0,0 +1,77 @@
...
+        .padding(.vertical, 12)
```
bgoncal: Use DesignSystem.Spaces

Sources/Extensions/Widgets/LiveActivity/HALockScreenView.swift
```diff
@@ -0,0 +1,77 @@
...
+        if let iconSlug = state.icon {
+            // UIColor(hex:) from Shared handles CSS names and 3/6/8-digit hex; non-failable.
+            let uiColor = UIColor(hex: state.color ?? haBlueHex)
+            let mdiIcon = MaterialDesignIcons(serversideValueNamed: iconSlug)
+            Image(uiImage: mdiIcon.image(
+                ofSize: .init(width: 20, height: 20),
+                color: uiColor
+            ))
+            .resizable()
+            .frame(width: 20, height: 20)
```
bgoncal: width and height are duplicated, create a constant to hold them

Sources/Extensions/Widgets/LiveActivity/HALockScreenView.swift
```diff
@@ -0,0 +1,77 @@
...
+    /// Parse hex color from ContentState, fallback to Home Assistant blue.
+    private var accentColor: Color {
+        Color(hex: state.color ?? haBlueHex)
+    }
+}
+
+// MARK: - Constants
+
+/// Home Assistant brand blue — used as fallback for icon and progress bar tints.
+let haBlueHex = "#03A9F4"
```
bgoncal: HA blue already exists in colors assets Color.haPrimary

Sources/Extensions/Widgets/Widgets.swift
```diff
@@ -34,6 +37,9 @@ struct WidgetsBundle17: WidgetBundle {
     }
 
     var body: some Widget {
+        if #available(iOSApplicationExtension 16.2, *) {
+            HALiveActivityConfiguration()
+        }
```
bgoncal: Is this check needed? since the bundle is iOS 17+ already?

Sources/Extensions/Widgets/Widgets.swift
```diff
@@ -54,6 +60,9 @@ struct WidgetsBundle18: WidgetBundle {
     }
 
     var body: some Widget {
+        // Live Activities (ActivityKit requires iOS 16.2+, this bundle requires iOS 18.0+)
```
bgoncal: This comment does not bring context needed

Sources/Shared/LiveActivity/HALiveActivityAttributes.swift
```diff
@@ -0,0 +1,109 @@
...
+/// ActivityAttributes for Home Assistant Live Activities.
+///
+/// Field names intentionally mirror the Android companion app's notification fields
+/// so that automations can target both platforms with minimal differences.
+///
+/// ⚠️ NEVER rename this struct or its fields post-ship.
+/// The `attributes-type` string in APNs push-to-start payloads must exactly match
+/// the Swift struct name (case-sensitive). Renaming breaks all in-flight activities.
+@available(iOS 16.2, *)
+public struct HALiveActivityAttributes: ActivityAttributes {
```
bgoncal, line 14: Can a test be created to validate this?

Sources/Shared/LiveActivity/HALiveActivityAttributes.swift
```diff
@@ -0,0 +1,109 @@
...
+        /// Hex color string for icon accent. Maps to `notification_icon_color`.
+        public var color: String?
+
+        // MARK: - Computed helpers (not sent over wire)
+
+        /// Progress as a fraction in [0, 1] for use in SwiftUI ProgressView.
+        public var progressFraction: Double? {
+            guard let p = progress, let m = progressMax, m > 0 else { return nil }
+            return Double(p) / Double(m)
+        }
```
bgoncal, line 63: Are these attributes intended to be generic for future live activities or use case specific for the progress bar activity that is being proposed now?

Sources/Shared/LiveActivity/LiveActivityRegistry.swift
```diff
@@ -0,0 +1,319 @@
...
+    /// Push-to-start (iOS 17.2+) allows HA to start a Live Activity entirely via APNs
+    /// without the app being in the foreground. This is best-effort (~50% success from
+    /// terminated state) — the primary flow remains notification command → app starts activity.
+    ///
+    /// The token is stored in Keychain and reported to HA via registration update so the
+    /// relay server can use it to send push-to-start APNs payloads.
+    ///
+    /// Call this once at app launch; the stream is infinite and self-managing.
+    @available(iOS 17.2, *)
+    public func startObservingPushToStartToken() async {
```
bgoncal, line 245: Based on that then we should make live activities feature itself iOS 17.2+ dependent right? And not iOS 16 as stated above.

Sources/Shared/LiveActivity/LiveActivityRegistry.swift
```diff
@@ -0,0 +1,319 @@
...
+
+    // MARK: - Public Helpers
+
+    /// The stored push-to-start token for inclusion in registration payloads.
+    /// Returns nil if the device hasn't received a token yet (pre-iOS 17.2 or not yet issued).
+    public static var storedPushToStartToken: String? {
+        AppConstants.Keychain[pushToStartTokenKeychainKey]
+    }
+
+    static let pushToStartTokenKeychainKey = "live_activity_push_to_start_token"
```
bgoncal, line 268: Can you create a test to validate this value wont change?

Sources/Shared/LiveActivity/LiveActivityRegistry.swift
```diff
@@ -0,0 +1,319 @@
...
+    /// The token is used by the relay server to send APNs updates directly to this activity.
+    private func reportPushToken(_ tokenHex: String, activityID: String) async {
+        let request = WebhookRequest(
+            type: "mobile_app_live_activity_token",
+            data: [
+                "activity_id": activityID,
+                "push_token": tokenHex,
+                "apns_environment": Current.apnsEnvironment,
+            ]
+        )
```
bgoncal: Can you create a test to validate this dictionary wont change?

Sources/Shared/LiveActivity/LiveActivityRegistry.swift
```diff
@@ -0,0 +1,319 @@
...
+    /// Notify HA servers that the Live Activity was dismissed or ended externally.
+    /// This allows HA to stop sending APNs updates for this activity.
+    private func reportActivityDismissed(activityID: String, tag: String, reason: String) async {
+        let request = WebhookRequest(
+            type: "mobile_app_live_activity_dismissed",
+            data: [
+                "activity_id": activityID,
+                "tag": tag,
+                "reason": reason,
+            ]
```
bgoncal: same here

Sources/Shared/Notifications/NotificationCommands/HandlerLiveActivity.swift
```diff
@@ -0,0 +1,179 @@
...
+    /// a local notification would silently fail if notification permission is not granted.
+    private static func showPrivacyDisclosureIfNeeded() {
+        guard !Current.settingsStore.hasSeenLiveActivityDisclosure else { return }
+        Current.settingsStore.hasSeenLiveActivityDisclosure = true
+    }
+
+    // MARK: - Validation
+
+    static func isValidTag(_ tag: String) -> Bool {
+        guard tag.count <= 64 else { return false }
```
bgoncal, line 98: Can you add a comment explaining the check?

Sources/Shared/Notifications/NotificationCommands/NotificationsCommandManager.swift
```diff
@@ -22,6 +22,12 @@ public class NotificationCommandManager {
         register(command: "clear_notification", handler: HandlerClearNotification())
         #if os(iOS)
         register(command: "update_complications", handler: HandlerUpdateComplications())
+        #if canImport(ActivityKit)
+        if #available(iOS 16.2, *) {
+            register(command: "live_activity", handler: HandlerStartOrUpdateLiveActivity())
+            register(command: "end_live_activity", handler: HandlerEndLiveActivity())
```
bgoncal, line 28: Can you create a test to validate these values wont change?

Sources/Shared/Notifications/NotificationCommands/NotificationsCommandManager.swift
```diff
@@ -36,8 +42,21 @@ public class NotificationCommandManager {
...
+        guard let hadict = payload["homeassistant"] as? [String: Any] else {
+            return .init(error: CommandError.notCommand)
+        }
+
+        // Support data.live_activity: true as an alternative to message: live_activity.
+        // This allows the notification body to be a real message instead of a command keyword,
+        // matching Android's data.live_update: true pattern.
+        #if canImport(ActivityKit)
+        if #available(iOS 16.2, *), hadict["live_activity"] as? Bool == true,
+           let handler = commands["live_activity"] {
```
bgoncal, line 53: same here

Sources/Shared/Settings/SettingsStore.swift
```diff
@@ -220,6 +220,17 @@ public class SettingsStore {
...
 
+    /// Whether the one-time Live Activity lock screen privacy disclosure has been shown.
+    /// Set to true after the first Live Activity is started; never reset.
+    public var hasSeenLiveActivityDisclosure: Bool {
+        get {
+            prefs.bool(forKey: "hasSeenLiveActivityDisclosure")
+        }
+        set {
+            prefs.set(newValue, forKey: "hasSeenLiveActivityDisclosure")
+        }
```
bgoncal, line 231: Sometimes on my iPhone I see the propmpt to allow live activities again even if I allowed them before, will this be reflected here as well?

Sources/Extensions/Widgets/LiveActivity/HADynamicIslandView.swift
```diff
@@ -0,0 +1,143 @@
...
+            HAExpandedBottomView(state: state)
+                .padding(.horizontal, 8)
+                .padding(.bottom, 4)
+        }
+    } compactLeading: {
+        HADynamicIslandIconView(slug: state.icon, color: state.color, size: 16)
+            .padding(.leading, 4)
+    } compactTrailing: {
+        HACompactTrailingView(state: state)
+            .padding(.trailing, 4)
```
bgoncal: Use DesignSystem.Spaces

Sources/Extensions/Widgets/LiveActivity/HADynamicIslandView.swift
```diff
@@ -0,0 +1,143 @@
...
+                .font(.caption2)
+                .foregroundStyle(.white)
+                .lineLimit(1)
+                .frame(maxWidth: 50)
+        } else if let fraction = state.progressFraction {
+            Text("\(Int(fraction * 100))%")
+                .font(.caption2)
+                .foregroundStyle(.white)
+                .monospacedDigit()
+        }
```
bgoncal, line 104: maxWidth is now duplicated, make a constant for them

Sources/Extensions/Widgets/LiveActivity/HADynamicIslandView.swift
```diff
@@ -0,0 +1,143 @@
...
+}
+
+// MARK: - Expanded bottom
+
+@available(iOS 16.2, *)
+struct HAExpandedBottomView: View {
+    let state: HALiveActivityAttributes.ContentState
+
+    var body: some View {
+        VStack(spacing: 4) {
```
bgoncal: Use DesignSystem.Spaces

Sources/Extensions/Widgets/LiveActivity/HADynamicIslandView.swift
```diff
@@ -0,0 +1,143 @@
...
+            if state.chronometer == true, let end = state.countdownEnd {
+                Text(timerInterval: Date.now ... end, countsDown: true)
+                    .font(.body.monospacedDigit())
+                    .foregroundStyle(.white)
+            } else {
+                Text(state.message)
+                    .font(.subheadline)
+                    .foregroundStyle(.white.opacity(0.85))
```
bgoncal, line 145: Use color from colors asset catalog (or create one if needed)

Sources/Extensions/Widgets/LiveActivity/HALiveActivityConfiguration.swift
```diff
@@ -0,0 +1,20 @@
...
+
+@available(iOS 16.2, *)
+struct HALiveActivityConfiguration: Widget {
+    var body: some WidgetConfiguration {
+        ActivityConfiguration(for: HALiveActivityAttributes.self) { context in
+            HALockScreenView(
+                attributes: context.attributes,
+                state: context.state
+            )
+            .activityBackgroundTint(Color.black.opacity(0.75))
```
bgoncal: Use color from colors asset catalog (or create one if needed)

bgoncal:
In your example we have 2 different tags, one for android and one for iOS:
```
    live_update: true           # Android 16+
    live_activity: true         # iOS 16.2+
```

Should we use the same as android has so the user does not need 2 different approaches based on platform? Are there more differences that we could merge into a single approach?

Sources/Extensions/Widgets/LiveActivity/HALockScreenView.swift
rwarner: Fixed — using DesignSystem.Spaces.one

Sources/Extensions/Widgets/LiveActivity/HALockScreenView.swift
rwarner: Extracted to private static let secondaryWhite constant

Sources/Extensions/Widgets/LiveActivity/HALockScreenView.swift
rwarner: Same — uses secondaryWhite constant now

Sources/Extensions/Widgets/LiveActivity/HALockScreenView.swift
rwarner: Fixed — DesignSystem.Spaces.two

Sources/Extensions/Widgets/LiveActivity/HALockScreenView.swift
rwarner: Fixed — DesignSystem.Spaces.oneAndHalf

Sources/Extensions/Widgets/LiveActivity/HALockScreenView.swift
rwarner: Created private static let iconSize: CGFloat = 20 — used in both ofSize and .frame()

Sources/Extensions/Widgets/LiveActivity/HALockScreenView.swift
rwarner: Replaced with Color.haPrimary fallback. haBlueHex scoped to private static where UIColor(hex:) still needs it

Sources/Extensions/Widgets/Widgets.swift
rwarner: Good call. Removed

Sources/Extensions/Widgets/Widgets.swift
rwarner: Removed

Sources/Shared/Notifications/NotificationCommands/HandlerLiveActivity.swift
rwarner: Added doc comment explaining allowed characters ([a-zA-Z0-9_-], max 64) and why

Sources/Extensions/Widgets/LiveActivity/HADynamicIslandView.swift
rwarner: Fixed — DesignSystem.Spaces.half

Sources/Extensions/Widgets/LiveActivity/HADynamicIslandView.swift
rwarner: Created private static let compactTrailingMaxWidth: CGFloat = 50

Sources/Extensions/Widgets/LiveActivity/HADynamicIslandView.swift
rwarner: Fixed — DesignSystem.Spaces.half

Sources/Extensions/Widgets/LiveActivity/HADynamicIslandView.swift
rwarner: Uses Color.haPrimary fallback via private accentColor computed property

Sources/Extensions/Widgets/LiveActivity/HALiveActivityConfiguration.swift
rwarner: Extracted to private static let lockScreenBackground

Sources/Shared/LiveActivity/HALiveActivityAttributes.swift
rwarner: Added testAttributesTypeName_isFrozen() — validates String(describing:) matches "HALiveActivityAttributes". Also testContentState_codingKeys_areFrozen() validates all JSON keys match Android field names, and testContentState_roundTrip_preservesAllFields() ensures no data loss through encoding.

Sources/Shared/LiveActivity/LiveActivityRegistry.swift
rwarner: Added testPushToStartTokenKeychainKey_isFrozen() — asserts the exact string value.

Sources/Shared/Notifications/NotificationCommands/NotificationsCommandManager.swift
rwarner: Added testLiveActivityCommandStrings_areFrozen() — validates both "live_activity" and "end_live_activity" commands route successfully.

Sources/Shared/Notifications/NotificationCommands/NotificationsCommandManager.swift
rwarner: Added testLiveActivityDataFlag_isRecognized() — validates live_activity: true flag is recognized.

Sources/Shared/LiveActivity/HALiveActivityAttributes.swift
rwarner: Generic — designed to cover the most common real-time monitoring use cases (progress bars, countdowns, status text, icons). The field names mirror Android's Live Notifications API so one YAML automation targets both platforms. New optional fields can be added without breaking existing activities (CodingKeys are additive-safe). Use-case-specific layouts would be a future enhancement on the widget view side, not the data model.

Sources/Shared/LiveActivity/LiveActivityRegistry.swift
rwarner: The core Live Activity feature (start, update, end via local push) works on iOS 16.2+. Push-to-start is 17.2+ only and is guarded separately with #available(iOS 17.2, *). Gating the entire feature at 17.2 would exclude users on iOS 16.2–17.1 who can still use Live Activities via local push. However, totally understand if that is a big downfall and pointless if it only works via local push. Open to what you think

Sources/Shared/LiveActivity/LiveActivityRegistry.swift
rwarner: Extracted webhookTypeToken and tokenWebhookKeys as static constants on LiveActivityRegistry. Added testWebhookTypeToken_isFrozen() and testTokenWebhookKeys_areFrozen() in LiveActivityContractTests.

Sources/Shared/LiveActivity/LiveActivityRegistry.swift
rwarner: Same — extracted webhookTypeDismissed and dismissedWebhookKeys. Added testWebhookTypeDismissed_isFrozen() and testDismissedWebhookKeys_areFrozen().

Sources/Shared/Settings/SettingsStore.swift
rwarner: This flag only tracks whether we've shown the in-app privacy disclosure text (in the Settings screen), not the actual iOS permission state. The real permission is always checked live via
ActivityAuthorizationInfo().areActivitiesEnabled before starting any activity — if iOS re-prompts and the user denies, that returns false and we skip the Activity.request() call.

Unless you mean there's a bug or something or you've seen this for other apps.

rwarner:
> > Local WebSocket — HA instance compiled with my core changes on same network
> 
> You have that already right? Because you opened a PR for core as well, so you can use that instance you used to test core to connect to the iOS App

Correct yeah, I haven't stood up core before locally from the repo will take a stab and see if I can ping HASS Core directly from the localhost from the iPhone sim. Was just using my personal instance to try at first.

--------------------

> In your example we have 2 different tags, one for android and one for iOS:
> 
> ```
>     live_update: true           # Android 16+
>     live_activity: true         # iOS 16.2+
> ```
> 
> Should we use the same as android has so the user does not need 2 different approaches based on platform? Are there more differences that we could merge into a single approach?

Great point actually. I can change this to just use the pre-existing `live_update` YAML that Android already uses and nobody has to make YAML additions / changes. (This would also require updating the other repos but honestly might be worth it)

Unless you think it would be handy to have the following example:

> they might want a Live Activity on iOS but a regular sticky notification on Android (or vice versa). 

But overall, the field names inside data (tag, progress, chronometer, etc.) are already unified across both platforms — it's only the opt-in trigger that's separate.

--------------------

Responded to all comments and questions, updated suggested code changes and integrated requested tests.

rwarner:
Great news, I was able to get core up and running locally and run the web socket test. Found two bugs I pushed up involving using the simulator. Got the live activity coming in via YAML:

https://github.com/user-attachments/assets/ac917cf1-5f6b-4e0f-9bf3-9fdc4897af63

bgoncal:
@rwarner so cool! Thanks for the extra effort on this. Allow me some days to validate the solution, I will be jumping between your PRs as soon as I can, I'm currently stuck on solving a bug related to the recently introduced mtls connection (testflight)

rwarner:
I had a feeling something was up with the dynamic island and the simulator. I went to work on another side project of mine and got it all working over there for the dynamic island and wanted to bring my findings over here. Now the dynamic island is working amazingly. I actually really like it. Just pushed up a rebase, some copilot fixes, and these dynamic island fixes. Here's some screenshots and a video of it in action (for each of the debug test buttons)


https://github.com/user-attachments/assets/a2bb874c-b37d-4e13-b088-da97c253a230


-----
**Irrigation**
Screenshot 2026-03-25 at 9 23 40 AM
Screenshot 2026-03-25 at 9 23 54 AM

-----
**Alarm**
Screenshot 2026-03-25 at 9 24 11 AM
Screenshot 2026-03-25 at 9 24 17 AM

-----
**All Fields**
Screenshot 2026-03-25 at 9 24 36 AM
Screenshot 2026-03-25 at 9 24 47 AM

-----
**Laundry**
Screenshot 2026-03-25 at 9 25 08 AM
Screenshot 2026-03-25 at 9 25 21 AM

-----
**EV Charging**
Screenshot 2026-03-25 at 9 25 37 AM
Screenshot 2026-03-25 at 9 25 42 AM

-----
**Music**
Screenshot 2026-03-25 at 9 26 11 AM
Screenshot 2026-03-25 at 9 26 16 AM

-----
**Shipments**
Screenshot 2026-03-25 at 9 26 36 AM

Screenshot 2026-03-25 at 9 26 39 AM

-----
**Camera**
Screenshot 2026-03-25 at 9 26 58 AM

bgoncal:
I'm manually testing the debug options you included and it looks great, I think we can improve the UI for the live activity itself in lock screen but this can be an iteration.

Can you add 2 things?
1 - Beta label to the live activities entry in settings (we have now a BetaLabel swiftUI view that you can reuse)
2 - Add a check Current.isTestFlight, to not let it slip into a release without we fully test it (after the PRs get merged)

Also have you checked how it behaves in macOS?

rwarner:
> I'm manually testing the debug options you included and it looks great, I think we can improve the UI for the live activity itself in lock screen but this can be an iteration.

Awesome. Sounds good

> Can you add 2 things? 1 - Beta label to the live activities entry in settings (we have now a BetaLabel swiftUI view that you can reuse) 2 - Add a check Current.isTestFlight, to not let it slip into a release without we fully test it (after the PRs get merged)

Added, let me know if this is right: https://github.com/home-assistant/iOS/pull/4444/changes/ff658536aef2b029f076063f86e54c69493e2d81

> Also have you checked how it behaves in macOS?

I have not, not familiar with running this directly in macOS. Let me know what I should look at in regard to that.

Also, I know we still gotta decide about the differentiating flags between iOS/Android

bgoncal:
> I have not, not familiar with running this directly in macOS. Let me know what I should look at in regard to that.

Maybe I am confusing how mac live activities work, I think they only display the ones coming from your iPhone, anyway, your checks for "supports live activity" should be enough for now.

> Also, I know we still gotta decide about the differentiating flags between iOS/Android

I missed where I commented about it but yes, I want android and iOS to have as minimum differences as possible, if one ignores tags that the other uses thats fine, but I dont want to duplicate tags that represent the same

rwarner:
> Maybe I am confusing how mac live activities work, I think they only display the ones coming from your iPhone, anyway, your checks for "supports live activity" should be enough for now.

No idea, not sure. Sounds good.

> I missed where I commented about it but yes, I want android and iOS to have as minimum differences as possible, if one ignores tags that the other uses thats fine, but I dont want to duplicate tags that represent the same

No problem, has grown into a large PR here lol

- Updated code for supporting `live_update` and removed `live_activity` for [core](https://github.com/home-assistant/core/pull/166072/commits/0aa42a1d1908af4b8c86f1a2a7d75fd4feb78fd7), [documentation](https://github.com/home-assistant/companion.home-assistant/pull/1303/commits/3815b4b5435ee2c1117ea24d092895c7d25b048b), [iOS](https://github.com/home-assistant/iOS/pull/4444/commits/f22b862c179637c8afcbb1c551ad047d84bfc4c0)
- Updated related PR descriptions
- Updated technical-brief.pdf

Let me know if we want to cutoff the feature for iOS 17+ rather than 16.2+ and I can do that too.

bgoncal:
I see many checks for iOS 16.2 but since we can only start activities from server side (Home Assistant) on 17.2+, doesnt it make more sense to support starting on 17.2?

EDIT: I see my comment overlapped yours haha sorry, yes let's move for 17

bgoncal:
Are you on discord? If so can you reach out to me, I want to have a quick chat about the full picture and If I understood your implementation correctly (bgoncal2)

rwarner:
- Updated live activity minimum to `17.2`
- Updated the PRs descriptions for this new minimum
- Regenerated technical-brief.pdf with this new minimum
- Updated codebase for [iOS](https://github.com/home-assistant/iOS/pull/4444/commits/10878eecd139eb2f6865b51be5920bf6b9e352b8) and [documentation](https://github.com/home-assistant/companion.home-assistant/pull/1303/commits/004ee90662a60a206c815106220b0eb5c1247e9b) for the new minimum

Yes I am, I'll reach out!

Sources/Shared/Environment/Environment.swift
```diff
@@ -130,7 +130,40 @@ public class AppEnvironment {
...
+            }
+            let registry = LiveActivityRegistry()
+            _liveActivityRegistryBacking = registry
+            return registry
+        }
+        set {
+            _liveActivityRegistryBacking = newValue
+        }
+    }
+    #endif
```
bgoncal: To avoid this custom approach you can make `liveActivityRegistry` optional, and since its a protocol, you can make the protocol available for all versions, but then when not on a supported OS you can just keep it `nil`

Sources/Shared/Environment/Environment.swift
```diff
@@ -156,7 +189,9 @@ public class AppEnvironment {
 
     public var cachedApis = [Identifier<Server>: HomeAssistantAPI]()
 
-    public var apis: [HomeAssistantAPI] { servers.all.compactMap(api(for:)) }
+    public var apis: [HomeAssistantAPI] {
+        servers.all.compactMap(api(for:))
+    }
```
bgoncal: Avoid making changes on what is not related to the PR main focus so it's easier to review and not worry that something that was working before may crash (I know it's just formatting in this case, but it speeds up the review process)

Sources/Shared/Environment/Environment.swift
```diff
@@ -258,6 +293,12 @@ public class AppEnvironment {
         #endif
     }()
 
+    /// Centralized gate for importing custom mTLS client certificates.
+    /// TestFlight-only for now; update this in one place if rollout rules change.
+    public var allowsCustomMTLSCertificateImport: Bool {
+        isTestFlight
```
bgoncal: This will conflict with a recent change, better to update from `main`

Sources/Shared/Environment/Environment.swift
rwarner: Great suggestion. Moved `@available(iOS 17.2, *)` to the individual protocol methods so the protocol itself is version-agnostic, then swapped the `Any?` backing for `LiveActivityRegistryProtocol?` with a runtime `#available` check in the getter. All call sites updated to `optional-chain`

Sources/Shared/Environment/Environment.swift
rwarner: Reverted

Sources/Shared/Environment/Environment.swift
```diff
@@ -130,7 +130,40 @@ public class AppEnvironment {
...
+                let registry = LiveActivityRegistry()
+                _liveActivityRegistryBacking = registry
+                return registry
+            }
+            return nil
+        }
+        set {
+            _liveActivityRegistryBacking = newValue
+        }
+    }
```
bgoncal: Why do we need both? Based on your comment one lazy var would be enough right?

rwarner:
> I think we can merge this one, let's just add a `Current.isTestFlight` check in the settings entry since we still need to validate that the integration with the FCM and Core PRs will work and not get this shipped by mistake.

Sounds good, I will add this in.

Definitely, I would hate to get anything merged in that doesn't work with the full end to end obviously outside of just the websocket working. In recent news, I did pay for the Apple Developer account so let me know if there's something I could potentially try on my end to test the FCM. Especially considering the minor changes I did yesterday for the `Core` comments that were left in that one

bgoncal:
You would need to create your own firebase project, link it in a local core installation and modify the app `.plist` to point to your firebase.

In FCM repo there are more instructions if you wanna give a try

rwarner:
> add a `Current.isTestFlight` check in the settings entry

SettingsItem.swift:155-158 and :165-167 already have Current.isTestFlight checks guarding the Live Activities entry. But let me know if this isn't correct

```
            // Live Activities require iOS 17.2+ and TestFlight
            if item == .liveActivities {
                if #available(iOS 17.2, *) { return Current.isTestFlight }
                return false
            }
```

```
    static var generalItems: [SettingsItem] {
        var items: [SettingsItem] = [.general, .gestures, .kiosk, .location, .notifications]
        if #available(iOS 17.2, *), Current.isTestFlight {
            items.append(.liveActivities)
        }
        return items
    }
```

If this is taken care of not sure if everyone is satisfied to give this a merge!

bgoncal:
Let's focus on the other PRs now ✅

PR #4469, opened 2026-04-02
Prevent execution of WKUserScript from beyond main frame and allowed hosts

bgoncal:

### DIFF
--- a/Sources/App/Frontend/WebView/WebViewController+Onboarding.swift
+++ b/Sources/App/Frontend/WebView/WebViewController+Onboarding.swift
@@ -7,19 +7,23 @@ import UIKit
 extension WebViewController {
     /// If user has not chosen 'Most secure' or 'Less secure' local access yet, this triggers a screen for decision
     func checkForLocalSecurityLevelDecisionNeeded() {
-        if Current.location.permissionStatus == .notDetermined, server.info.connection.hasNonHTTPSURLOption {
+        let connection = server.info.connection
+
+        if Current.location.permissionStatus == .notDetermined, connection.hasNonHTTPSURLOptions {
             Current.Log.verbose("User has not decided location permission yet")
             showOnboardingPermissions(steps: OnboardingPermissionsNavigationViewModel.StepID.updateLocationPermission)
-        } else if server.info.connection.connectionAccessSecurityLevel == .undefined {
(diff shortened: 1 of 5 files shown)


Sources/App/Frontend/ExternalMessageBus/SafeScriptMessageHandler.swift
```diff
@@ -13,8 +16,29 @@ final class SafeScriptMessageHandler: NSObject, WKScriptMessageHandler {
...
+        ) else {
+            return
+        }
         delegate?.userContentController(
             userContentController, didReceive: message
         )
     }
+
+    func shouldAllowMessage(isMainFrame: Bool, host: String) -> Bool {
+        isMainFrame && allowedHosts.contains(host)
```
TimoPtr, line 32: Can't you check the current URL loadded in the Webview and check if they have the same origin ?

Sources/App/Frontend/ExternalMessageBus/SafeScriptMessageHandler.swift
```diff
@@ -13,8 +16,29 @@ final class SafeScriptMessageHandler: NSObject, WKScriptMessageHandler {
...
+    }
+
+    private var allowedHosts: Set<String> {
+        let urls = [
+            server.info.connection.address(for: .internal),
+            server.info.connection.address(for: .external),
+            server.info.connection.address(for: .remoteUI),
+        ]
+
+        return Set(urls.compactMap { $0?.host })
```
TimoPtr, line 42: Does host contains the port on iOS? if not the origin should be HOST+PORT

Sources/App/Frontend/ExternalMessageBus/SafeScriptMessageHandler.swift
bgoncal: I could, but this would be problematic in case the attacker finds a way to place their website as the main frame, then I would be comparing to the malicious URL, which doesn't make sense, comparing to what is configured for the server is safer

Sources/App/Frontend/ExternalMessageBus/SafeScriptMessageHandler.swift
bgoncal: Good point, I'll add that

bgoncal: https://github.com/home-assistant/iOS/pull/4470

PR #4491, opened 2026-04-09
Add barometer as an entity that iPhone/some iPads expose

teancom:
## Summary
I was reading this article https://www.howtogeek.com/your-phone-has-a-powerful-weather-instrument-hidden-inside/ - and it seemed neat! And adding support was actually pretty simple. Mostly just the one file plus tests and a little scattering of changes to the Xcode project file.

Documentation: home-assistant/companion.home-assistant#1309

## Any other notes
Nothing bigger, just thought it would be neat.


bgoncal:
How's the permission request flow? I see it has "unauthorized" as one of the possible state

Sources/Shared/API/Webhook/Sensors/BarometerSensor.swift
```diff
@@ -0,0 +1,66 @@
...
+            return .init(error: BarometerError.unavailable)
+        }
+
+        let (promise, seal) = Promise<CMAltitudeData>.pending()
+        let queue = OperationQueue()
+        queue.name = "barometer-sensor"
+
+        Current.barometer.startUpdatesOnQueueHandler(queue) { data, error in
+            // We only need a single reading, so stop updates immediately
+            Current.barometer.stopUpdates()
```
bgoncal, line 124: Isn't this something that should update often?

teancom: I was following the pedometer sensor as a bit of a template, but I can absolutely switch over to SensorProviderUpdateSignaler like BatterySensor does and have it send more continuous updates if you would prefer that approach. Frankly, I'm just not sure what the impact is to battery life so I went with a conservative approach. If that's not an issue, great.

teancom:
> How's the permission request flow? I see it has "unauthorized" as one of the possible state

I took the same approach as PedometerSensor and ActivitySensor, which are also bundled into the Motion & Fitness permission but otherwise are silently skipped in the list of available sensors. I figured this was enough of a niche thing that it shouldn't be something that pops up and requests the user grant permission how we do for Bluetooth, Notifications, or Camera (etc).

Sources/Shared/API/Webhook/Sensors/BarometerSensor.swift
bgoncal: What are the use cases you have in mind for this sensor?

Then we can define how often we want to update it.

I would expect that who activates this sensor would expect it updating more than just once every long period of time

teancom: Yeah, I was probably being overly cautious. I was thinking about triggers on "big" drops in a short time as an alert source, which would require more frequent updates. I used to live in Nebraska and that was a sign to head for shelter.

I'll update the PR tonight to switch over.

teancom: Alright, that was a little trickier than I thought (async stuff can be tricky, who knew) but that has been working on my phone and should be good to go (🤞)

teancom:
I did what I should have done before and replicated the CI/CD environment locally to see what was really going on vs. guessing. Tracked it down, this *should* now be good to go (at least, the tests should pass cleanly).

bgoncal:
@teancom can you check the tests?

teancom:
Man, it was surprisingly difficult to figure out why everything passed locally and failed in CI. I'm hoping this is the final-for-real-fix-v1-(final) commit. 😅

PR #4422, opened 2026-03-09
Add kiosk mode core infrastructure (PR 1/5)

nstefanelli:
## Summary

Reopening kiosk mode PR (previously #4218) after addressing all review feedback from @bgoncal. Apologies for the delay — some family matters pulled me away, but I'm excited to get this implemented and am fully committed to seeing it through.

This PR adds the foundational infrastructure for **kiosk mode** — a feature designed for wall-mounted iPad displays running Home Assistant dashboards.

### What's included:
- **KioskModeManager** — Central coordinator singleton with KioskModeObserver protocol (no NotificationCenter)
- **KioskSettings** — Comprehensive settings model with GRDB persistence (camelCase columns, Codable)
- **Screensaver system** — Clock screensaver with 4 styles, pixel shift for OLED burn-in prevention
- **Secret exit gesture** — Configurable corner tap to access settings when locked down
- **Status bar hiding** — StatusBarForwardingNavigationController for proper UIKit integration
- **Settings UI** — SwiftUI settings accessible via Settings menu and secret gesture
- **28 unit tests** — Covering settings serialization, time logic, orientations, enums

### Changes since #4218 (all review feedback addressed):
- ✅ Fixed settings modal dismissal on iOS 15 (UINavigationController + explicit onDismiss closure)
- ✅ Fixed status bar hiding (StatusBarForwardingNavigationController + statusBarView integration)
- ✅ Removed non-functional settings (navigationLockdown, TouchFeedbackManager, IconMapper)
- ✅ Replaced `Date()` with `Current.date()` for testability
- ✅ GRDB persistence follows CarPlayConfig/WatchConfig pattern with DatabaseTableProtocol
- ✅ All strings localized via L10n (SwiftGen)
- ✅ SFSafeSymbols used throughout (no string-based systemName)
- ✅ KioskModeObserver protocol (no NotificationCenter)
- ✅ Kiosk files prefixed with `Kiosk`
- ✅ UIKit screensaver container (required for status bar/home indicator control)
- ✅ macCatalyst filtered out
- ✅ Merged with latest upstream/main (adapted to Frontend/ directory restructure)
- ✅ Idle timer pauses when settings view is open
- ✅ TODO comments (no PR-specific comments)

### Testing Instructions
1. **Settings → Companion App → Kiosk Mode** → Enable "Enable Kiosk Mode"
2. Set screensaver timeout to **30 seconds** for quick testing
3. Wait for screensaver → tap to wake
[...]

Issue #2403, opened 2023-08-29
iOS Companion App Not Executing Query String in URI


kirsch33:
**iOS device model, version and app version**

Model Name: iPhone 13 
Software Version:  iOS 16.6
App version: HA companion app version 2023.460

**Home Assistant Core Version**

Home Assistant server version 2023.8.4, supervisor 2023.08.1, operating system 10.5

**Describe the bug**

Using a script in home assistant, it is possible to open a Safari link directly into the HA app if the URI provided is a lovelace subdirectory. for example, this automation will open my Frigate camera dashboard within HA (correctly) when executed
```yaml
script:
  send_dynamic_notification:
    sequence:
      - service: notify.mobile_app_austins_iphone
        data:
          message: "{{ message }}"
          data:
            tag: "{{ notifTag }}"
            group: "{{ group }}"
            image: "{{ null }}"
            video: "{{ null }}"
            attachment:
              url: "{{ image }}"
            push:
              interruption-level: active
            actions:
              - action: URI
                title: "View Clip In Browser"
                uri: "{{ video }}"
                destructive: false
              - action: URI
                title: "View Live"
                uri: "/lovelace/frigate"
                destructive: false
    mode: single
```

However, the Frigate card (https://github.com/dermotduffy/frigate-hass-card) is capabale of passing actions thru the URI as well via query strings. however, HA companion app does not recognize or utilize any provided query strings.

in the above example, if I change this line:

```
uri: "/lovelace/frigate"
```

to

```
uri: "/lovelace/frigate?frigate-card-action:live&frigate-card-action:camera_select={{ camera }}"
```

this should open the specific requested camera, but does not. 

I have tested that this works fine on the iOS Safari app itself, but just not as intended with the HA companion app itself.

**Additional context**
For additional context see my post in this issue over on the frigate-hass-card repo 
https://github.com/dermotduffy/frigate-hass-card/issues/1253#issuecomment-1696566488

PR #4497, opened 2026-04-10
Kiosk mode: camera motion detection for screen wake

nstefanelli:
## Summary

Adds camera-based motion and presence detection to kiosk mode (PR2 of kiosk series, per [discussion #2403](https://github.com/orgs/home-assistant/discussions/2403#discussioncomment-16305655)).

- **Motion detection**: CoreImage pixel-diff at 5 fps with configurable sensitivity (low/medium/high)
- **Presence detection**: Vision framework person/face detection with hysteresis (2-frame confirm, 10s absence timeout)
- **Screen wake**: Configurable wake-on-motion and wake-on-presence triggers
- **Settings UI**: New "Camera Detection" section with toggles, sensitivity picker
- **macCatalyst**: Camera detection excluded on Mac

Builds on #4422 (PR1: core kiosk infrastructure).

## Architecture

```
KioskCameraDetectionManager (coordinator)
├── KioskCameraMotionDetector (CoreImage pixel-diff, 5 fps, .low preset)
└── KioskPresenceDetector (Vision person/face detection, 3-10 fps, .medium preset)
         │
         ▼
KioskModeManager.wakeScreen(source:)
```

- Frame processing runs on dedicated serial queues, not the main thread
- State updates dispatched to MainActor via Combine publishers
- Presence activity timer keeps screen awake while person is present

## New files

| File | Lines | Purpose |
|------|-------|---------|
| `Camera/KioskCameraDetectionManager.swift` | ~200 | Coordinator singleton |
| `Camera/KioskCameraMotionDetector.swift` | ~250 | CoreImage motion detection |
| `Camera/KioskPresenceDetector.swift` | ~300 | Vision presence/face detection |
| `KioskCameraDetection.test.swift` | ~115 | MotionSensitivity + settings tests |

## Test plan

- [ ] MotionSensitivity enum: threshold values, Codable roundtrip, ordering
- [ ] Camera settings: defaults, roundtrip, backwards compatibility with pre-camera JSON
- [ ] Enable motion detection → wave hand → screen wakes from screensaver
- [ ] Enable presence detection → walk up → screen wakes; walk away → screensaver resumes
- [ ] Deny camera permission → detection doesn't start, no crash
- [ ] Both detectors simultaneously → no conflict
- [ ] macCatalyst build → camera section not visible

## Related

- Discussion: https://github.com/orgs/home-assistant/discussions/2403
- PR1: #4422 (merged)
- L10n fix: #4490 (merged)


bgoncal:
Sorry the delay to pickup this review, I will still need a bit more time since I am working on something else right now, meanwhile I noticed that after releasing the foundation for kiosk mode, lots of questions about how it works and comparisons to the famous "Fully Kiosk browser" popped up in the community, so it would be nice if we created a documentation page for this in out companion app docs, could you propose it?

Reference: https://companion.home-assistant.io/docs/getting_started/

We need to make it clear things like... in iOS we cannot lock the user in a single app (from Home Assistant) and that the closest way possible to that is if the user uses iOS "guided access". So it needs to explain the difference of the "lock" that we mention inside Home Assistant app and the "lock" that the user may be expecting.

Macstar1601:
> Sorry the delay to pickup this review, I will still need a bit more time since I am working on something else right now, meanwhile I noticed that after releasing the foundation for kiosk mode, lots of questions about how it works and comparisons to the famous "Fully Kiosk browser" popped up in the community, so it would be nice if we created a documentation page for this in out companion app docs, could you propose it?
> 
> Reference: https://companion.home-assistant.io/docs/getting_started/
> 
> We need to make it clear things like... in iOS we cannot lock the user in a single app (from Home Assistant) and that the closest way possible to that is if the user uses iOS "guided access". So it needs to explain the difference of the "lock" that we mention inside Home Assistant app and the "lock" that the user may be expecting.

The Kiosk Mode feature is really great. But for me, it’s not important to be locked into Kiosk Mode. It’s much more important that the Dashboard can be accessed more quickly and that you don’t have to unlock the iPad first. A true lock within an app also leads to other problems, such as not being able to use other apps at the same time, e.g., intercom apps. I opened an issue about this today: https://github.com/home-assistant/iOS/issues/4506

Sources/App/Kiosk/Camera/KioskCameraDetectionManager.swift
```diff
@@ -0,0 +1,216 @@
...
+    @Published public private(set) var motionDetected: Bool = false
+
+    /// Current presence detected state
+    @Published public private(set) var presenceDetected: Bool = false
+
+    /// Current face detected state
+    @Published public private(set) var faceDetected: Bool = false
+
+    /// Number of faces detected
+    @Published public private(set) var faceCount: Int = 0
```
bgoncal: Should we start with motion and then make iteration PRs for face related detections? To keep the testing scope simpler per PR

bgoncal (changes requested):

bgoncal:
@Macstar1601 thanks for sharing, we dont need to worry about that because the app is not able to lock itself as the only app, iOS doesn't allow that, the maximum we can do is to prevent the screen to turn off.

nstefanelli:
Stripped face/presence so this PR is motion-only. The inline items are addressed in the latest commits. Companion-app docs proposal coming in a separate issue.

bgoncal:
Thanks for that, let me know whenever you have the docs PR so we can review that first before adding more features

bgoncal (changes requested):
Let's keep it in draft while we don't have the docs

bgoncal:
@nstefanelli Do you need some help with the docs? Or if you have created already and I missed it please tag me and I'll review asap

nstefanelli:
Thanks @bgoncal — opened the docs PR over at [home-assistant/companion.home-assistant#1323](https://github.com/home-assistant/companion.home-assistant/pull/1323). It covers kiosk mode end-to-end (PR1 + this PR), flagged as BETA, with screenshot placeholders left in for you to flag whichever captures you want included. Happy to mark this PR ready for review whenever the docs side looks good to you, or earlier if that's easier on your end.

nstefanelli:
Rebased on current main, ready for review @bgoncal!

Sources/App/Kiosk/KioskSettings.swift
```diff
@@ -117,6 +120,120 @@ public struct KioskSettings: Codable, Equatable {
...
+        case clockShowDate
+        case clockUse24HourFormat
+        case clockStyle
+        case secretExitGestureEnabled
+        case secretExitGestureCorner
+        case secretExitGestureTaps
+        case cameraMotionEnabled
+        case cameraMotionSensitivity
+        case wakeOnCameraMotion
+    }
```
bgoncal, line 161: Why did you have to add coding keys? you can choose the variable names so you could avoid that right?

bgoncal:
Apologies on the delay to test how this PR behaves on a real device, Xcode 26.5 is refusing to run the app on my iPad so I will use the distribution CI to test it internally and then I will post back the results here

bgoncal (changes requested):
One thing it's nice to add is a "Notify on motion" with a subtitle "Use this option to see a feedback when motion is detected, this can be helpful to determine the sensitivity needed."

You can use the "ToastManager" to display this visuals, if the minimum iOS version for the ToastManager is not enough then use LocalNotificationDispatcher.


Also can you please include screenshots? Including how it looks when the camera permission has been denied.

bgoncal:
I tested, the motion detection works but for some reason if I close and reopen the app kiosk mode is back disabled, can you check that as well? When fixing this please provide a video including closing and reopening the app so it can be quickly validated

bgoncal:
I'm closing this PR for now due to foundational issues related to Kiosk mode that need to be fixed before anything is added as disclosed here: https://github.com/OpenHomeFoundation/roadmap/issues/51#issuecomment-4705893547

Also, there was a recent refactor which moved a lot of the UI to SwiftUI so perhaps makes it easier to reimplement kiosk mode in a more organized way without linking everything to the WebViewController. Let's continue the discussion in the epic.

PR #4574, opened 2026-04-28
Enable Live Activities on iPad (iPadOS 17+)

rwarner:
## Summary

iPad has supported Live Activities since iPadOS 17 — the previous `userInterfaceIdiom == .pad` guard in `LiveActivitySettingsView` was incorrect. `areActivitiesEnabled` (the `ActivityAuthorizationInfo` API) already returns the correct value on iPad, so no platform-specific gate is needed.

Changes:

- Remove the `userInterfaceIdiom == .pad` branch from `LiveActivitySettingsView` that showed a static "Not available on iPad" label. iPad users now see the same enabled/open-settings states as iPhone.
- Drop the `live_activity.status.not_supported` localisation key from all 34 locale files and the generated `Strings.swift` accessor.
- Update a stale comment in `HAAPI.swift` that incorrectly stated iPads report `false` for `areActivitiesEnabled`.

ActivityKit handles Dynamic Island availability transparently — the island pill appears on iPhone 14 Pro/Pro Max and all iPhone 15 and later models; on iPad and older iPhones it simply doesn't render, and the activity shows on the Lock Screen only.

## References

- Apple docs confirming iPad support: [Displaying live data with Live Activities](https://developer.apple.com/documentation/activitykit/displaying-live-data-with-live-activities)
- Companion docs PR: home-assistant/companion.home-assistant#1303

## Test plan

- [ ] On iPhone: Live Activities settings screen shows enabled/open-settings as before
- [ ] On iPad (iPadOS 17+): settings screen no longer shows "Not available on iPad"; shows enabled/open-settings
- [ ] On iPad (iPadOS 17+): Live Activities can be started and updated from Home Assistant
- [ ] On iPad (iPadOS 16 or earlier): `areActivitiesEnabled` returns `false`; open-settings button shown
- [ ] Verify `live_activity.status.not_supported` key is absent from all `.lproj` files

### DIFF
--- a/HomeAssistant.xcodeproj/project.pbxproj
+++ b/HomeAssistant.xcodeproj/project.pbxproj
@@ -11436,7 +11436,7 @@
 			isa = XCBuildConfiguration;
 			baseConfigurationReference = 9C4E5E25229D986B0044C8EC /* HomeAssistant.beta.xcconfig */;
 			buildSettings = {
-				DEVELOPMENT_TEAM = QMQYCKL255;
+				DEVELOPMENT_TEAM = 4Q9RHLUX47;
 			};
 			name = Beta;
 		};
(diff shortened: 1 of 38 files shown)


bgoncal (approved):

bgoncal:
I manually deleted `live_activity.status.not_supported` in Lokalise ✅

PR #4604, opened 2026-05-04
Add confirmation popup for tags URL handling + screen to manage allowed URLs

bgoncal:
## Summary

This PR adds a prompt to prevent tag URLs to be executed directly without user previous knowledge.

TDB: Copy

## Screenshots

CleanShot 2026-05-04 at 15 03 19@2x


bgoncal:
@TimoPtr @jpelgrom drafted the tag url misuse prevention, copy to be defined and updated, let me know your thoughts

jpelgrom:
Having 'Tags' as a settings menu item in companion app settings could be confusing. Now there is Settings > Tags and Settings > Companion app > Tags. I understand why (the list is also for non-NFC tags) and don't have a good alternative right now.

How does text overflow? The tag ID is already close to touching the right edge of the container, if it hides overflow it may not be possible to distinguish different tags starting with the same ID.

In the database you appear to be saving server IDs as well. Not sure if this is a quirk specific to how your database works or intentional. If you want to link it to server IDs this creates an issue: the UI suggests a URL is always allowed, while it may not be allowed if servers were added/removed.

TimoPtr:
> Having 'Tags' as a settings menu item in companion app settings could be confusing. Now there is Settings > Tags and Settings > Companion app > Tags. I understand why (the list is also for non-NFC tags) and don't have a good alternative right now.
> 
> How does text overflow? The tag ID is already close to touching the right edge of the container, if it hides overflow it may not be possible to distinguish different tags starting with the same ID.
> 
> In the database you appear to be saving server IDs as well. Not sure if this is a quirk specific to how your database works or intentional. If you want to link it to server IDs this creates an issue: the UI suggests a URL is always allowed, while it may not be allowed if servers were added/removed.

I'm not a big fan of adding the tags settings because it already exists in the frontend. We could just in the troubleshooting section add a button revoke all granted tag? Or add a button within the frontend itself?

PR #4660, opened 2026-05-24
Add Assist VAD timing settings

lextiz:
Adds Assist VAD silence and timeout settings to iOS voice input.


bgoncal:
This should be controlled solely in core, iOS does not need specific controllers for VAD.
So if the feature is desired please use Core to centralize the implementation so its assist pipeline-specific

PR #4665, opened 2026-05-26
Add "in_zones" array when reporting zone-only location

bgoncal:
## Summary

To fulfill the new requirements in core we need to send in_zones when we send a location update using zone only. See https://github.com/home-assistant/architecture/discussions/1387 for more details. Implemented in core https://github.com/home-assistant/core/pull/171814

Android PR: https://github.com/home-assistant/android/pull/6879


Sources/Shared/API/HAAPI.swift
```diff
@@ -666,11 +666,14 @@ public class HomeAssistantAPI {
             update = .init(trigger: updateType, location: rawLocation, zone: zone)
             location = rawLocation
         case .zoneOnly:
+            let supportsInZones = server.info.version >= .inZonesOnLocationUpdate
             if updateType == .BeaconRegionEnter {
-                update = .init(trigger: updateType, usingNameOf: zone)
+                let zones = zone.flatMap { $0.TrackingEnabled && !$0.isPassive ? [$0] : nil } ?? []
```
TimoPtr: Core is stripping the passive zone in the logic but I think we should still then them, to be sure that core has all the information we have.

bgoncal: Sorry I didn't understand, you want to remove the filter and send them all? I would be fine with that, perhaps @emontnemery should pitch in?

emontnemery: Please send all zones to core.
Core's `in_zones` state attribute includes both passive and active zones.
Core will pick the name of the smallest active zone as state (with fallback to "not home" if there's no active zone in in_zones).

PR #4671, opened 2026-05-27
Fix remote Live Activity token lifecycle

bgoncal:
## Summary

- Observe ActivityKit activities that are started remotely via APNs push-to-start notifications.
- Register the stored push-to-start token under `live_activity_token`, matching Core's expected app data key.
- Send Live Activity token and dismissal webhooks using Core's strict schema.
- Update Live Activity contract tests for the frozen webhook wire format.

## Root Cause

APNs could start the Live Activity, but the app was not observing `Activity<HALiveActivityAttributes>.activityUpdates`. That meant the app did not attach its normal per-activity push-token and lifecycle observers to remotely-started activities, so Core could not receive the per-activity push token needed for subsequent updates.


rwarner:
Hey @bgoncal, found a small unrelated bug while testing on the simulator, `clear_notification` was being silently dropped on simulator/Catalyst (Direct interface). Different from the real-device issue this PR + [core#166072](https://github.com/home-assistant/core/pull/166072#pullrequestreview-4372386873) addresses. `NotificationManagerLocalPushInterfaceDirect` never assigns `localPushManager.delegate`, so silent commands get dropped. iOS doesn't fire `willPresent` for them and the delegate is the only fallback. The Extension path on real iPhone already handles this at line 197.

I didn't know if you'd rather fold this into this PR rather than making a whole nother one for it. Up to you:

One-line fix in `NotificationManagerLocalPushInterfaceDirect.init:`

```swift
self.localPushManagers = .init { [weak self] server in
    let manager = LocalPushManager(server: server)
    manager.delegate = self?.localPushDelegate   // ← add this
    let token = NotificationCenter.default.addObserver(
```

Verified it clearing the notification on the simulator with the local push websocket on top of all these changes and my incoming `core` changes.

rwarner:
Also found a second related bug. The `countdownEnd: Date?` in `HALiveActivityAttributes.ContentState` is auto-Codable, so ActivityKit decodes it with JSONDecoder's `default.deferredToDate` strategy (seconds since 2001-01-01). HA core emits Unix epoch seconds (and the in-app handler already constructs `Date(timeIntervalSince1970:)`), so the APNs-path decode would land ~31 years in the future and the countdown timer would render wrong. The in-app path bypasses this so simulator local-push doesn't hit it.

Custom init(from:) + encode(to:) on ContentState reading the value as Double then Date(timeIntervalSince1970:) 

Branch: https://github.com/rwarner/iOS/tree/fix/live-activity-codable-and-direct-delegate (the second commit, ea5067be2). Same question for if you want to include it in your PR here or have it be seperate

bgoncal:
@rwarner Please open a PR for those fixes you mentioned above, I think it makes sense, even though I dislike have custom decoders, if there is no other way, thats fine

PR #4739, opened 2026-06-13
Fix notification/deep-link URL being discarded on cold start (#4145)

### LINKED ISSUE #4145: Home Assistant Companion App ignores url parameter
Model Name: iPhone 15 Pro Max
Software Version: iOS 26.1
App version: Home Assistant Companion 2025.12.1 (2025.1546)

Home Assistant OS Core version: 2025.12.4

If the Home Assistant Companion app is already open, the notification test shown below works; however, if the Home Assistant Companion app is not already running, it just opens the app.  The url parameter is discarded.

```
action: notify.all_phones
data:
  title: "Cold Start Attempt"
  message: "Force close app, then tap."
  data:
    url: "http://host_ip_goes_here:8123/frigate?index=/review/review_id_goes_here"
    tag: "cold_start_test"
```

In case it matters somehow, I am using Frigate 0.16.3; but this does not seem to be a Frigate issue.  I am using an unauthenticated (port 5000) access point on the remote Frigate server.  As mentioned earlier, this works well if the Home Assistant Companion app is already open when the notification gets clicked.

dimdal:
## Summary

Fixes #4145. Tapping a notification with a `url` (or a `URI` action targeting the Home Assistant frontend) navigates correctly when the app is already running, but on a **cold start** the URL is discarded and the app just opens to its default page.

**Root cause — a cold-start race.** On cold start the webview starts blank (`webView.url == nil`). The notification tap drives `open(inline:)` → `load(notificationURL)`, but `loadActiveURLIfNeeded()` — fired independently from `viewWillAppear` and `HomeAssistantAPI.didConnectNotification` — loads the **default** server URL whenever `webView.url == nil`. Whichever `load()` lands last wins, and it is usually the default, so the notification URL is lost. When the app is already running `webView.url` is non-nil, `loadActiveURLIfNeeded()` early-returns, and `open(inline:)` always wins — which matches the reports (works when open, and one reporter seeing it work intermittently is the race surfacing).

**Fix.** Mirror the existing `initialURL` restoration mechanism with a `pendingOpenInlineURL`:
- `open(inline:)` records the explicitly requested URL before loading it.
- `loadActiveURLIfNeeded()` gives that pending URL priority over the restored/default URL (new pure helper `prioritizedInlineURL(...)`).
- `didFinish` clears it once the navigation lands.

Because the pending URL is re-asserted while the webview is still blank and only cleared on a successful navigation, it wins in **every** ordering of the two loads, so the URL is no longer discarded. Deep links benefit too, since they share the same `open(inline:)` path.

Scope: this covers URLs that resolve to the HA frontend (server URLs and relative paths such as `/frigate/review/...`), which is the path in the issue. The separate `action: URI` → external browser path (`openURLInBrowser`) is not touched here.

Added unit tests in `WebViewControllerTests` for the priority helper and for `open(inline:)` recording the pending URL.

## Screenshots

No UI change — this is a navigation/timing fix, so there is nothing visual to capture. Behaviour: with the app force-closed, tapping a notification carrying a `url` now lands on that URL instead of the default dashboard.

## Any other notes

### DIFF
--- a/HomeAssistant.xcodeproj/project.pbxproj
+++ b/HomeAssistant.xcodeproj/project.pbxproj
@@ -1588,6 +1588,7 @@
 		E60778133EC44B10A61D551C /* TestFlightCommunicationModels.swift in Sources */ = {isa = PBXBuildFile; fileRef = 212E30CB76DE4ED1B6E317CD /* TestFlightCommunicationModels.swift */; };
 		E92E09E3A93650D56E3C5093 /* KioskScreensaverViewController.swift in Sources */ = {isa = PBXBuildFile; fileRef = 62CDCFDB29D283A7902A3ABE /* KioskScreensaverViewController.swift */; };
 		F0D1DD41A8F55F6D767EBF37 /* TemplatePreviewSection.swift in Sources */ = {isa = PBXBuildFile; fileRef = 332C060487B468055C487070 /* TemplatePreviewSection.swift */; };
+		F810A6B2F8EE83B572364BBB /* WebViewControllerPendingURLTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 851BAEAD0BF400015BB5FE9F /* WebViewControllerPendingURLTests.swift */; };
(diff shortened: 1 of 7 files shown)


dimdal:
Superseded by #4740 — the head branch was renamed (dropping a personal prefix), which closed this PR. Same commits and fix; continuing in #4740.

PR #4779, opened 2026-06-18
[WIP] Xcode 27

### LINKED ISSUE #4749: Xcode 27 no longer supports watchOS 8 deployment target
### Issue

Xcode 27 beta no longer supports watchOS 8 as a deployment target. When building the Home Assistant iOS project with Xcode 27 beta, the watchOS target fails with:

```text
The watchOS deployment target 'WATCHOS_DEPLOYMENT_TARGET' is set to 8.0, but the range of supported deployment target versions is 9.0 to 27.0.x.
```

Right now, HA supports watchOS 8 and newer, so the project still uses a watchOS 8 deployment target.

Apple’s Xcode system requirements list Xcode 27 beta as supporting watchOS deployment targets 9–27:
https://developer.apple.com/xcode/system-requirements/

### Proposed solution

Unless there is another way around it, we need to raise the Home Assistant watchOS deployment target from 8.0 to 9.0. This would drop Apple Watch Series 3 support, since Series 3 is limited to watchOS 8. 

### Notes

This also affects normal iOS app builds/archives, not only building the watchOS target directly. The iOS app includes an embedded Watch app, so the Watch app must build successfully as part of the app build/archive process. With the embedded Watch app still configured for a watchOS 8 deployment target, Xcode 27 beta fails before the full iOS app build can complete.

bgoncal:


bgoncal:
`main` can already compile on xcode 27

PR #4781, opened 2026-06-18
Apply the partial-fill gauge style to the Lock Screen (`accessoryCircular`)

bahoo:
## Summary

Follow-up to #4741, which introduced the partial-fill Gauge on the Home Screen (`systemSmall`). This applies the same treatment to the Lock Screen (`accessoryCircular`).

The `.normal` and `.singleLabel` gauge types now render the colored `0→value` arc over a dim track — matching the Home Screen and Apple's Batteries gauge — instead of the full ring + indicator dot. The arc inherits the system's monochrome tint, so the bright-fill / dim-track band survives the Lock Screen's vibrant rendering. `.capacity` stays native on both surfaces (it already fills `0→value` itself).

The arc is inset to the native accessory footprint on the Lock Screen so it stays within the system's circular safe area (it otherwise fills edge-to-edge and would be clipped) and stays consistent in size with the `.capacity` style.

Adds `accessoryCircular` snapshot coverage (normal / single-label / capacity) and `#Preview` blocks for both families. Verified on a device Lock Screen: the fill/track band reads in monochrome, nothing clips, and the labels stay legible.

## Screenshots

band_test

## Any other notes

Split out from #4741 — the Lock Screen change was deferred there as separate scope. The Home Screen rendering is unchanged by this PR.


bgoncal:
Simulator Screenshot - Daily tester 2 - 2026-06-22 at 10 10 21
minimum and max range numbers are too small.

In my opinion we dont need the new deisgn for the lock screen how it renders nowadays is ok, wdyt?

There is an area you could work on that would help a lot the users, add a new Gauge type called "Entity", user picks an entity and it automatically renders the UI, without templating yaml etc.

bgoncal:
Also please use full simulator screenshots when sharing the result in GitHub, for example, in your screenshot the labels looked fine, but as you can see on mine, they are too small, It's confusing

bgoncal:
I made some UI adjustments as well https://github.com/home-assistant/iOS/pull/4798

bahoo:
> my opinion we don't need the new deisgn for the lock screen how it renders nowadays is ok, wdyt?

I wanted the filled track effect for visual consistency in iOS ecosystem ( battery gauges, rings, etc. ) — but ...

> I made some UI adjustments as well #4798

Oh yeah, #4798 is looking gorgeous 😍 you've got cleaner styles and implementation both, here. Ship it. 🐿️ I will close this PR, I don't think there's anything else here to run with.

> There is an area you could work on that would help a lot the users, add a new Gauge type called "Entity", user picks an entity and it automatically renders the UI, without templating yaml etc.

I love this idea, though if the widget has to auto-infer / derive which values to show 🤔 I would want to get more acquainted with empirical conventions before taking a crack at it. Maybe start with a feature request ?

> Also please use full simulator screenshots when sharing the result in GitHub, for example, in your screenshot the labels looked fine, but as you can see on mine, they are too small, It's confusing

Got it. 👍🏻 Will do! Thanks!

PR #4787, opened 2026-06-19
Fix macOS menu-bar icon not activating the main window

bgoncal:
## Summary

When the macOS app is set to display in the **menu bar** (`.accessory` activation policy) and the main window has been closed, tapping the Home Assistant menu-bar icon did nothing — the `WindowGroup` was never activated and the UI stayed hidden.

Tapping the menu-bar item runs `Current.sceneManager.activateAnyScene(for: .webView)`, which calls `requestSceneSessionActivation` with a `userActivity` whose `targetContentIdentifier` is `"ha.webview"`. As documented in `SceneActivity.swift`, that identifier is what SwiftUI's `handlesExternalEvents(matching:)` matches against to route an activated scene to the correct `WindowGroup`.

The primary webView `WindowGroup` in `HAApp` was the **only** scene missing that modifier — Settings, About, Assist, and Onboarding all declare it (which is why opening those from the menu worked). With no group advertising `"ha.webview"`, SwiftUI couldn't bind the newly-requested scene to a `WindowGroup`, so no window appeared. Plain app launch worked because that scene carries no identifier and falls through to the default group; only the identifier-tagged re-activation hit the gap.

The fix adds `.handlesExternalEvents(matching: [SceneActivity.webView.activityIdentifier])` to the primary `WindowGroup`, consistent with the other four. Window reuse is still handled separately by `existingScenes(for:)` passing the existing session, so this won't spawn duplicate windows.

## Any other notes
Regression introduced in the SwiftUI app-lifecycle migration (#4748).


bgoncal:
Superseded by #4789, which combines this with #4788 and adds the actual fix for the icon not showing the window (the `handlesExternalEvents` matcher here was necessary for the closed-window case but not sufficient on its own).

PR #4802, opened 2026-06-22
Receive live activity tap and open correct server + accept url from push payload

bgoncal:


bgoncal:
@copilot address failing CI

PR #4853, opened 2026-06-26
Tell Home Assistant how long to wait for a Live Activity "update push token"

bgoncal:
## Summary

- This PR is in combination with core PR:
- It sets how long core should wait for the live activity push-token (the one used to update a live activity, not the one that starts it)
- This PR also increases the retry to communicate the push-token when it fails


bgoncal:
Closing the PR since alternative solutions were found using APNS headers in firebase (apns-collapse-id) + short background session to remove duplicates on iOS side.

PR #4941, opened 2026-07-05
Auto-switch server based on location

bgoncal:
## Summary
Adds opt-in location-based automatic server switching — the same behavior Apple Home uses to pick the current home. When enabled, the app decides which server to display at launch and on every return to the foreground, based on where the user currently is:

1. **Wi-Fi network**: a server whose `internalSSIDs` (or `internalHardwareAddresses` on macOS) contain the current network is preferred — the same "internal network" signal `ConnectionInfo.activeURL()` already uses.
2. **Zones**: otherwise, a fresh one-shot location is tested against each server's tracked (non-passive) zones, falling back to the cached `inRegion` state maintained by region monitoring when location is unavailable.

A switch only happens when a *single* server unambiguously matches — if two servers share an SSID or overlapping zones, the app stays where it is.

**Manual switches win for a while**: any intentional server change (server picker, next/previous-server gestures, deep links, notification taps — everything routing through `AppCoordinator.open(server:)`) is recorded, and the app will not switch away from that server for a 15-minute grace period (`LocationBasedServerSwitcherImpl.defaultGracePeriod`). So backgrounding and quickly returning doesn't yank the user back to the location-based server; after the grace period elapses, the next foreground/launch returns to it. The grace period persists across app restarts.

Implementation:
- New `LocationBasedServerSwitcher` protocol + implementation in `Sources/Shared/Location`, exposed as `Current.locationBasedServerSwitcher` (testable via the World pattern; system lookups are injectable).
- `OnboardingStateObservable` consults it for the initial screen and re-evaluates on `willEnterForegroundNotification`, on setting changes, and once after launch (the launch decision is synchronous and can only use cached state — the SSID is fetched asynchronously).
- New `SettingsStore.locationBasedServerSwitchEnabled` (default **off**) with a toggle in Settings under the servers list, shown only when more than one server is configured. Kiosk mode always takes precedence over auto-switching.
- Unit tests for SSID matching, zone matching (fresh location + cached fallback), ambiguity handling, enablement gating, and the grace-period lifecycle.

## Screenshots

The new "Auto-switch server" toggle appears in Settings directly below the Servers section (only with 2+ servers configured). Screenshots to be added — this change was authored in an environment without an iOS simulator.

## Any other notes
- The decision logic never picks a server arbitrarily: no unambiguous match → no switch.
- The one-shot location request is bounded to a few seconds so a foreground evaluation can't stall; it falls back to the region-monitoring zone state.
- The 15-minute grace period is a constant that's easy to tune or surface as a setting later if desired.

https://claude.ai/code/session_01MW4B6WyzJwfENSGTWzE88C

---
_Generated by [Claude Code](https://claude.ai/code/session_01MW4B6WyzJwfENSGTWzE88C)_


bgoncal:
@copilot resolve the merge conflicts in this pull request

PR #5039, opened 2026-07-11
Migrate background refresh task to modern BGTaskScheduler

bgoncal:


hiSandog:
One lifecycle edge case worth testing is scheduling after the app is relaunched while a previously submitted refresh request is still pending. If the scheduler rejects duplicate requests, the code should treat that differently from a genuine submission failure and still reschedule after each completed or expired task. It would also be helpful to verify that every asynchronous exit path calls `setTaskCompleted` exactly once, including expiration racing with the refresh operation’s normal completion.

bgoncal:
@hiSandog thanks for the heads up, I will run some tests over the following week(s), can you perhaps help me testing this? Not sure your background but, if you are also a dev, that would be very apreciated.

PR #5079, opened 2026-07-13
Split complication settings views, add floating live preview

bgoncal:
## Summary

- **One view per file**: `ComplicationListView.swift` held three view structs; `ComplicationsRootView` and `WatchComplicationBuilderEditView` now live in their own files (code moved verbatim), and the legacy list gained its missing `#Preview`. The one-type-per-file + `#Preview` rule is now surfaced in a "Non-Negotiable Conventions" section of `AGENTS.md`.
- **Previews work without onboarding**: `FakeServerManager` is now `public`, and the builder's `#Preview` seeds `Current.servers` with it, so the server-gated form (including the live complication mock) renders in Xcode previews.
- **Floating live preview**: new reusable `FloatingPanel` in `HADesignSystem` (registered in the components gallery) — a picture-in-picture panel that follows the finger, snaps to the nearest corner on release (flick-aware via predicted end point), and resizes by tap-toggle or pinch. In the complication builder, once the inline preview section scrolls off-screen the preview re-appears floating over the form, showing only the selected family centered and zoomed to fit a compact watch face; the panel chrome radius is concentric with the watch bezel.
- **Customize disclosure persists**: the editor's "Customize" toggle is stored in a new nullable `isCustomized` column on `WatchComplicationConfig` (added automatically by the existing column migration); rows saved before the flag fall back to inferring it from per-size overrides.

CleanShot 2026-07-13 at 23 39 03@2x


Sources/Shared/Database/GRDB+Initialization.swift
```diff
@@ -206,6 +206,8 @@ final class WatchComplicationConfigTable: DatabaseTableProtocol {
                     t.column(DatabaseTables.WatchComplicationConfig.customGaugeTemplate.rawValue, .text)
                     t.column(DatabaseTables.WatchComplicationConfig.sortOrder.rawValue, .integer).notNull()
                     t.column(DatabaseTables.WatchComplicationConfig.families.rawValue, .jsonText)
+                    // Nullable: absent falls back to inferring from `families` (see showsCustomized()).
```
bgoncal:
```suggestion
```

PR #5098, opened 2026-07-14
Improve watch-iPhone communication reliability

bgoncal:
## Summary
Six reliability fixes for the WatchConnectivity layer, from an audit of the watch↔iPhone communication. One commit per fix:

1. **Assist audio chunk transfer** — chunks now send sequentially with ack-driven backpressure instead of being fired all at once with no error handler. A lost chunk surfaces as an error instead of a silent hang, the completion callback fires on success (it never did), and each recording carries a unique `recordingId` so a retried recording can never mix stale chunks from the previous attempt into corrupt audio. The phone drops partial uploads idle for 60s instead of leaking them.
2. **Database sync buffer leak** — the phone's in-memory chunk buffer was only freed when the watch requested the final chunk; abandoned syncs leaked the whole encoded mirror. A new sync start now frees previous buffers, with a 10-minute expiry backstop.
3. **`encodeForWatch` no longer crashes** — it called `fatalError` on encode failure; it now throws and all senders fall back gracefully (keep local copy / reply treated as decode failure by the watch, never as an authoritative empty config).
4. **Guaranteed-message hygiene** — every reload while unreachable queued another `transferUserInfo` config pull and the phone answered them all; duplicates are now skipped via the session's outstanding transfers. The phone's guaranteed reply also spells out `GuaranteedMessage` explicitly instead of compiling as it by overload-resolution luck.
5. **Application-context echo-back removed** — each side sent the counterpart's keys back (watch battery bounced back to the watch, complication payloads back to the phone), doubling every update toward the 262 KB cap and able to resurrect stale values. Each side now sends only the keys it owns.
6. **Payload-size guards** — the ~65 KB `sendMessage` and ~262 KB application-context ceilings were only a comment. Every outgoing message and reply is now measured, oversized ones are logged with their identifier and byte count, and an oversized context is rerouted through the database-mirror `transferFile` push (no size cap) instead of failing.

## Screenshots
No UI changes; behavior changes are in transport reliability (previously-hanging assist sessions now surface errors).

## Any other notes
Stacked on #5097 (the `HAWatchCommunicationMessages` package extraction) — the first commit here is that PR's commit, and this diff reduces to the six fixes once it merges. All watch-related unit tests pass (43/43 in Tests-Shared + Tests-App). The wire protocol stays compatible with older counterpart builds: the `recordingId` is optional on the phone side, and all identifiers/raw values are unchanged.

### DIFF
--- a/Sources/HAWatchCommunicationMessages/Sources/InteractiveImmediateResponses.swift
+++ b/Sources/HAWatchCommunicationMessages/Sources/InteractiveImmediateResponses.swift
@@ -8,6 +8,10 @@ public enum InteractiveImmediateResponses: String, CaseIterable {
     case pushActionResponse = "PushActionResponse"
     case assistPipelinesFetchResponse
     case assistAudioDataResponse
+    /// Phone → watch: per-chunk acknowledgement of `assistAudioDataChunked`, carrying
+    /// `{acknowledged, chunkIndex, totalChunks}`. The watch sends the next chunk only after
+    /// receiving this, so audio streams with backpressure instead of flooding the session.
+    case assistAudioChunkAck
     case assistSTTResponse
     case assistIntentEndResponse
     case assistTTSResponse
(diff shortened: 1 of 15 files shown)


bgoncal:
@copilot fix lint issues reported by CI

PR #4954, opened 2026-07-06
Migrate remaining Realm models to GRDB

bgoncal:
## Summary
Moves the last remaining Realm models — zones (`RLMZone` → `AppZone`), notification categories/actions, watch complications and location history/errors — to GRDB, and migrates users' existing data before Realm is retired. After this PR, `import RealmSwift` exists in exactly one file: `RealmToGRDBMigration.swift`, a one-time importer that reads the legacy Realm store (running the historic schema migrations for very old stores) and writes the rows into GRDB on first launch of the app and the watch app. Once the migration has shipped for a few releases, deleting that file plus the `pod 'RealmSwift'` entry removes the dependency entirely — the pod is kept for now solely so user-authored data (locally created notification categories/actions, watch complications, per-zone settings) is not lost on update.

Highlights, in commit/review order:
1. **GRDB model layer** — `AppZone`, `NotificationCategory`/`NotificationAction` (actions embedded as JSON in the category row), `WatchComplication` (keeps its ObjectMapper wire format for phone↔watch context sync), `LocationHistoryEntry`/`LocationError`, plus five new tables registered in `GRDB+Initialization`. `LegacyModelManager` keeps its public surface but now diffs/stores/cleans up via GRDB, and observes via `ValueObservation`.
2. **Data migration** — `RealmToGRDBMigration` with private legacy `Object` mirrors (`@objc` names match the historic class names); retries on next launch if the import fails, and logs a `ClientEvent` with imported counts.
3. **Shared consumers** — HAAPI location submission, watch context sync, complication template rendering, geocoder sensor, one-shot location errors; `Current.realm`/`realmFatalPresentation` removed from the environment.
4. **iOS app target** — ZoneManager region monitoring driven by a zone `ValueObservation` (`.immediate` scheduling preserves the monitor-at-startup behavior), notification category editors reworked around value semantics, `RealmResultsObserver` → `DatabaseResultsObserver`, location history views, watch complication view models, developer tools.
5. **Watch/widget/CarPlay targets** — watch complication storage and sync, stale imports.
6. **Tests** — fixtures move from in-memory Realms to in-memory `DatabaseQueue`s; ZoneManager tests wait for the async observation where Realm delivered synchronously on `refresh()`.
7. **Xcode project** — file renames and removal of a dangling `ObjectMapper+RealmList.swift` reference (it was not compiled into any target).

_Generated by [Claude Code](https://claude.ai/code/session_01WjzY3ZsTdqA44cEq9PdrjQ)_

PR #5118, opened 2026-07-15
Restore model manager subscription so zones sync again

bgoncal:
## Summary

Zone enter/exit tracking is currently dead: the `appZone` GRDB table is never populated from the server, so `ZoneManager` has nothing to monitor and the Location settings screen shows no zones.

The states-cache subscription that keeps `AppZone` in sync (`LegacyModelManager.SubscribeDefinition.defaults` → `.states(domain: "zone", type: AppZone.self)`) is only started by `modelManager.subscribe(isAppInForeground:)`. That call lived in `WebViewSceneDelegate` and was dropped in the UIKit→SwiftUI app migration (#4748) without a replacement — since then, nothing in the app starts it.

The regression went unnoticed because stale `RLMZone` rows written before #4748 persisted on disk, so region monitoring kept working off old data. It became visible once the Realm→GRDB migration (#4954, fixed in #5116) switched reads to the fresh `appZone` table.

## Changes

- Call `Current.modelManager.cleanup()` and `Current.modelManager.subscribe(isAppInForeground:)` from `AppDelegate.setupModels()`, after the Realm→GRDB import and before `ZoneManager` is created. `LegacyModelManager` re-subscribes itself on server changes, so the single call at launch is sufficient.

## Testing

Verified in the simulator with two connected servers: on launch the logs show `store(AppZone)` creating all 22 zones (`from(0) eligible(21) new(21)` + `zone.home` on the second server) and `ZoneManager sync > available 22, monitoring 20, started 19`, where before it was `available 0, monitoring 0`. The `appZone` table (previously confirmed empty via sqlite) is now populated and zone data live-updates.

### DIFF
--- a/Sources/App/AppDelegate.swift
+++ b/Sources/App/AppDelegate.swift
@@ -425,6 +425,13 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
         // Import any legacy Realm data into GRDB before anything reads it
         RealmToGRDBMigration.migrateIfNeeded()
         NotificationCategory.setupObserver()
+        // Start the server-state subscriptions that keep GRDB models in sync
+        // (zones via the states cache); without this, appZone is never populated
+        // and region monitoring has nothing to track.
+        Current.modelManager.cleanup().cauterize()
+        Current.modelManager.subscribe(isAppInForeground: {
+            UIApplication.shared.applicationState == .active
+        })
     }
 
     private func setupMenus() {


nijaobscvra666:
Tested this on TestFlight beta 2026.7.2 (build 2026.2446, well after this commit was merged and distributed) and zones still aren't populating for me. Pulled the app's debug log to check what's actually happening rather than guessing.

`ZoneManager.sync(zones:)` shows `available 0` on every single app launch - I force-quit and relaunched 7 times over about 40 minutes, same result every time:

```
[Info] [main] [ZoneManager.swift:215] sync(zones:) > available 0, enabled true, monitoring 0 ((beacon: 0, circular: 0, zone: 0)), started 0, ended 0
```

More specifically, `LegacyModelManager.store(type:from:sourceModels:)` only ever logs updates for `NotificationCategory` across all 7 launches - `AppZone` never shows up in the log at all:

```
[Verbose] [...] [LegacyModelManager.swift:337] store(type:from:sourceModels:) > updating NotificationCategory server(...) from(0) eligible(0) deleted(0) ignored(0) new(0)
```

If this were purely the `isAppInForeground` timing issue flagged in review (`applicationState == .active` vs `!= .background`), I'd expect the whole subscription batch to succeed or fail together depending on app state at launch - not one type reliably syncing while another reliably doesn't, every single time. Since `NotificationCategory` comes through fine on every launch but `AppZone` never does, it looks like there might be something specific to the zone-domain states subscription itself, separate from (or in addition to) the foreground-check race.

Ruled out a couple of other things in the same log that looked suspicious at first but turned out unrelated:
- `GetMobileAppConfig()` returns 404 on every launch, but the app already handles this gracefully (`"ios component is not loaded; pretending there's no push config"`) - that's the old deprecated `ios` component, not `mobile_app`, expected to 404 on any modern instance.
- `AppDatabaseUpdater` logs a device registry parse error (`Failed to fetch device registry: ... incorrectType ... "identifiers"`), but that's a separate pipeline (Entities → Entities Registry → Entities persist → Devices Registry → Areas) that completes successfully regardless and has no zones step at all.

Happy to share more of the log if useful - trying to avoid dumping the whole 1.4MB file here.

bgoncal:
This change is not on 2026.7.2, it will be on 2026.8.0

PR #5272, opened 2026-07-28
Add Apple Health sensors

bgoncal:
## AI Policy

## Summary
The HealthKit sensor only reported step count and resting heart rate. This replaces those with a data-driven catalog of 22 Apple Health metrics, grouped into the categories the Health app uses: Activity, Body Measurements, Heart, Nutrition, Respiratory and Vitals.

The set is chosen to match what people actually ask for: everything the Android companion already reports through Health Connect that HealthKit also has, plus the metrics named in the iOS feature requests — heart rate, blood glucose, distance and weight. The catalog is plain data, so more can be added a category at a time; `HealthKitMetric+All.swift` lists what was left out and what adding one back involves.

Each metric declares its HealthKit identifier, unit, aggregation (daily total or newest sample), lookback window, rounding, and the device and state classes to register with. Identifiers are stored as raw strings so metrics from newer versions of iOS can be listed unconditionally — the service skips whatever the running OS doesn't provide, and checks the unit and aggregation style before querying.

Step count is dropped, since `PedometerSensor` already reports it from CoreMotion.

Enablement is an allowlist (#5273), so none of these report until the user switches one on. They get their own searchable screen rather than the main sensor list, which they would otherwise bury, and `SensorRegistry` is updated to list the catalog's IDs.

The permission request only asks for the sensors that are turned on. HealthKit never re-prompts for a type it has already been asked about, so asking again is what picks up sensors enabled since the last request — there's a button for that on the Apple Health sensor screen, next to the enable-all toggle, and tapping Apple Health under Permissions does the same rather than opening system settings, where Health has no switches.

Not included yet, and the most obvious follow-up: **sleep**. It's a HealthKit category type rather than a quantity type, so it needs a different query and state shape than everything here.

## Screenshots
_To add: Apple Health sensors screen in light and dark mode._

## Any other notes
The catalog is covered by tests that resolve every identifier against HealthKit on the test runner's OS and assert the declared unit and aggregation style match the real quantity type, so a wrong entry fails instead of silently reporting nothing.


aero-oli:
@bgoncal I noticed you’ve removed the HealthKit steps because we already get them from Core Motion.

I actually wonder if it might be better the other way around. As I understand it, Core Motion only reports steps detected by the iPhone’s own sensors, whereas HealthKit provides the consolidated step count, including data from an Apple Watch. That means HealthKit is generally a more accurate representation of the user’s actual step count.

bgoncal:
@aero-oli Oh, that makes sense, I will add back the "Health steps" then, thanks for the heads up!

PR #5319, opened 2026-08-01
Add per-server custom request headers

### LINKED ISSUE #5305: Support per-server custom headers for Cloudflare Access / Zero Trust gateways
## Feature request

Please consider adding optional per-server custom HTTP headers in the iOS Companion App, with secure storage, so users can place Home Assistant behind Cloudflare Access or similar Zero Trust gateways without weakening the Access policy.

This is related to the older closed request:

- #439

## Use case

Many Home Assistant users expose their instance through Cloudflare Tunnel + Cloudflare Access instead of opening a direct public port. For non-browser clients, Cloudflare Access supports service-token authentication through headers such as:

- `CF-Access-Client-Id`
- `CF-Access-Client-Secret`

The iOS Companion App currently does not appear to provide a way to attach these headers to all Home Assistant requests, including:

- onboarding/login flows;
- normal REST calls;
- WebSocket calls;
- frontend/webview asset requests if applicable.

## Security impact

Without header support, users who want the iOS app to keep working often have to choose a weaker Cloudflare Access setup, for example:

- bypassing Access for a country/IP range;
- exposing a second less-protected hostname for mobile use;
- relying only on Home Assistant authentication at the origin;
- using ad-hoc reverse-proxy exceptions.

That is worse than allowing the app to store a small set of explicit headers securely and send them only to the configured Home Assistant server URL.

In my own setup, I can protect other remote tools with Cloudflare Access service-token headers, but for Home Assistant iOS compatibility I have to keep a bypass policy. Native header support would let me remove that bypass and make the public Home Assistant endpoint stricter.

## Suggested behavior

Add an advanced per-server setting such as "Additional request headers" or "Gateway headers":

- stored securely in Keychain;
- scoped only to the configured Home Assistant base URL and its WebSocket URL;
- applied consistently to REST, WebSocket, and login/onboarding requests;
- not sent to arbitrary third-party URLs loaded by the app/webview;
- never logged or included in diagnostics;
- clearly marked as an advanced reverse-proxy / Zero Trust option.

Example configuration:

```text
CF-Access-Client-Id: <client id>
CF-Access-Client-Secret: <client secret>
```

Cloudflare also supports sending the service token as a single custom header, so a generic header mechanism would help more than only hard-coding Cloudflare-specific names.

## Alternatives considered

- Cloudflare Access browser login: works in browsers but is not reliable for all companion app/native/WebSocket flows.
- Access bypass policy: works, but reduces the value of using Cloudflare Access.
- mTLS: useful for some deployments, but service-token headers are widely used in Cloudflare Access and are easier for many users to rotate/manage.
- Separate mobile-only hostname: works, but increases operational complexity and can still require weaker policy.

## Why this helps Home Assistant users

This would let users keep Home Assistant behind a stronger Zero Trust policy while still using the iOS Companion App. It reduces the need for broad bypass rules and makes it easier to run a secure remote Home Assistant deployment without sacrificing mobile functionality.

tigercraft4:
## Summary

Adds per-server additional request headers for Home Assistant connections. This is intended for deployments behind reverse proxies or Zero Trust gateways such as Cloudflare Access service-token policies, where the iOS app needs to send headers like `CF-Access-Client-Id` and `CF-Access-Client-Secret` to reach the configured Home Assistant URL.

Fixes #5305.

## What changed

- Adds `AdditionalRequestHeader` to `ConnectionInfo`, persisted with the full server info.
- Adds validation/sanitization for header names and values:
  - allows normal HTTP token header names, including Cloudflare Access headers;
  - rejects reserved/sensitive transport or auth headers such as `Authorization`, `Cookie`, `Host`, `Content-Length`, `Content-Type`, `User-Agent`, and hop-by-hop headers;
  - rejects empty values, newline values, invalid names, and duplicate names case-insensitively.
- Applies additional headers only when the request URL matches one of the server's configured URLs by scheme/host/port and configured base path.
- Does not apply these headers to cloudhook URLs or unrelated hosts.
- Applies headers to:
  - Alamofire requests through `ServerRequestAdapter`;
  - token refresh/revoke routes;
  - initial token exchange when the onboarding `ConnectionInfo` already has headers;
  - HAKit's server-specific `URLSession` configuration;
  - top-level `WKWebView` navigations through the central `load(request:)` path.
- Adds a Connection Settings section for editing per-server headers, with values entered through `SecureField`.
- Strips additional request headers from `ServerInfo.mirroredForPersistence` so they are not copied into the sanitized GRDB mirror used for recovery.

## Security notes

The app still owns normal Home Assistant authentication. User-configured headers are intentionally blocked from overriding `Authorization`, cookies, host routing, content length/type, user agent, and hop-by-hop behavior. Headers are scoped to the configured server base URLs to avoid sending them to unrelated domains.

Header values are sensitive. They remain in the full server configuration, which is already stored with the rest of the server credential state, but are removed from the sanitized mirror.

## Known limitations / maintainer validation needed

`WKWebView` reliably allows custom headers on explicit top-level navigations. WebKit does not provide a general way to force custom headers onto every subresource request or every WebSocket created by frontend JavaScript. This PR covers the app-controlled navigation path and native REST/token paths; maintainers should validate whether Home Assistant frontend-created WebSocket/subresource requests inherit enough context for Cloudflare Access in supported deployments, or whether a deeper follow-up is required.

The full `Tests-Unit` scheme could not be run locally because this Mac does not have the watchOS 26.5 runtime installed, and the scheme builds an embedded Watch app.

## Tests

- `git diff --check`
- `xcodebuild build -project HomeAssistant.xcodeproj -scheme Shared-iOS -destination 'generic/platform=iOS Simulator' -derivedDataPath /tmp/ha-ios-dd-shared2 -skipMacroValidation -skipPackagePluginValidation CODE_SIGNING_ALLOWED=NO`

Additional attempted checks that were environment-blocked:

- `swift test` in `Sources/HANetworking`: blocked before compilation by package platform resolution (`HANetworking`/`HAModels` macOS 10.13 vs dependencies requiring newer macOS).
- `xcodebuild test -scheme Tests-Unit`: blocked by missing watchOS 26.5 runtime.
- `xcodebuild build -scheme App-Debug`: blocked by missing watchOS 26.5 runtime.


tigercraft4:
Addressed the Copilot review feedback in 3bd86fadd:\n\n- Restricted additional request header names to ASCII RFC token characters.\n- Debounced custom header persistence from the settings UI and flush pending edits on disappear to avoid Keychain writes on every keystroke.\n- Removed config-wide injection into makeCertificateAwareURLSession so custom gateway headers are not placed on a shared URLSessionConfiguration. Existing per-request/server-scoped paths continue to apply headers via URLRequest mutation/adapters.\n- Added coverage for rejecting non-ASCII header names and for not placing custom headers in the certificate-aware session configuration.\n\nValidation:\n- git diff --check\n- xcodebuild build -project HomeAssistant.xcodeproj -scheme Shared-iOS -destination 'generic/platform=iOS Simulator' -derivedDataPath /tmp/ha-ios-dd-shared-copilot -skipMacroValidation -skipPackagePluginValidation CODE_SIGNING_ALLOWED=NO\n\nTests-Unit is still blocked locally by the missing watchOS 26.5 runtime before test compilation starts. A direct Tests-Shared target build is also not useful here because target-only builds default to Release/iphoneos and hit an existing dependency cycle.

bgoncal:
Companion apps are currently aiming to center their security around mTLS, at the moment custom headers are not something that will be supported, each connectivity layer that we include in the project increases its maintainability across iOS, watchOS, Mac catalyst and specially on apps extensions such as widgets, shortcuts, watch complications etc, which we currently don't have bandwidth to maintain.

The contribution is appreciated but as of this moment it can't be added to the project, i recommend looking into client certificates as going forward, it adds true encryption and prevent unauthorized reachability of your server.

PR #5488, opened 2026-08-22
Fix custom widget state cache dropping tiles after the first 3

### LINKED ISSUE #3520: Custom widget not showing state for all entities after changing it.
**Device model, version and app version**

Model Name: MacBook Pro 16-inch, Nov 2023
macOS Version: Sequoia 15.2
App Version: 2025.2
Safari Version: 18.2

**Home Assistant Core Version** 2025.3.4

I created a custom widget and it has 11 entities listed in there. It is updating regularly however only the 1st 3 entities are showing the state, none of the others are showing the state. 
When I added a sensor widget from Home Assistant app with all the same entities it showed the state in there for all of them. And interesting enough my custom widget then also suddenly started to show the state for all entities and not just the 1st 3.

It seems to happen when editing the custom widget. As it was fine again after I had added the other widget, then I went to change the names displayed and from the moment I saved it updated the widget resulting in again only the 1st 3 being shown.

Removing the widget, clicking Done, and then re-adding it ensures again that all of them are shown. But then an update results again in the same issue however when I retried this it didn't work.
If I click "Reload all Widgets" then it shows values again for all the entities. Reload has to happen 1-2 minutes after saving otherwise it does not seem to do it. :-)

I had it run for a while (about 40 minutes) with no issues. Then made a change to the widget and from the moment I saved it happened. Clicked reload and didn't change. Waited a minute and then tried again and now it shows all of them again.

**To Reproduce**
Create a custom widget and add it to desktop. Then make updates to the widget and save it.

**Expected behavior**
State values for all entities are shown in the widget, not just the 1st 3.

**Screenshots**

Image showing custom widget with only values for the 1st 3 entities

Image

Image showing that with standard sensor widget as well the custom widget is now showing all values.
Image

**Additional context**

fengxguo:
## Summary
Custom widgets only showed HA state on the first 3 tiles after a save/reload because the small-family timeline (max 3 items) overwrote a shared widget-id cache that medium/large then reused.

## Changes
- Treat the states cache as valid only if it contains every requested item
- Merge newly fetched states into the existing cache instead of replacing it

Fixes #3520

## Test plan
- Custom widget, 4+ items, Show states ON, medium or large
- Edit the widget and save
- Confirm tiles 4+ still show state without waiting or adding a sensor widget

### DIFF
--- a/Sources/Extensions/Widgets/Common/WidgetTimelineProviderSupport.swift
+++ b/Sources/Extensions/Widgets/Common/WidgetTimelineProviderSupport.swift
@@ -91,16 +91,21 @@ struct WidgetEntityStateProvider {
             return [:]
         }
 
-        if let cache = readCache(), cache.cacheCreatedDate.timeIntervalSinceNow > -cacheValiditySeconds {
+        let neededItems = items.filter(itemFilter)
+        if let cache = readCache(),
+           cache.cacheCreatedDate.timeIntervalSinceNow > -cacheValiditySeconds,
+           Set(neededItems).isSubset(of: Set(cache.states.keys)) {
             Current.Log.verbose("\(logPrefix) widget states cache is still valid, returning cached states")
             return cache.states
         }
 
         Current.Log.verbose("\(logPrefix) widget has no valid cache, fetching states")
 
-        var states: [MagicItem: WidgetEntityState] = [:]
+        // Merge so a small-family refresh (prefix of 3) cannot wipe states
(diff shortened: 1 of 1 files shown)


fengxguo:
Closing — not pursuing this contribution.

bgoncal (changes requested):
Please reopen the PR following the PR template.

I'd like to reinforce that PRs must be handled by humans and not AIs directly.

fengxguo:
Closing — not pursuing this contribution.

PR #5492, opened 2026-08-22
Make the disconnected-state server picker tappable immediately

### LINKED ISSUE #5198: Pop-up button to change servers is unusable for two minutes upon launch, if previous URL of Home Assistant is unreachable
Model Name: iPhone 12 mini
Software Version: iOS 18.7.8
App version: 2026.7.1 (2026.2346)

2026.7.1

**Describe the bug**
I have one instance of Home Assistant running on my Raspberry Pi.  When I'm on my home network, I use a local IP address in the companion app to access it.  However, when I'm out and about, that local IP address obviously doesn't work.  I use the free version of Tailscale on my Raspberry Pi and on my iPhone to establish a connection, and use the Tailscale IP address to access Home Assistant when I'm _not_ on my home network.  (Yes, I could use Home Assistant cloud, but I'm using Tailscale for other things anyway.)

The only hiccup to this setup is that I have to set up two "servers" in the Home Assistant iOS companion app, even though they ultimately point to the same instance of Home Assistant.  I have a "Home" server set up, using my local network IP address URL ( http://raspberrypi.local:8123 ).  And then I have a second "Tailscale" server set up, using the Tailscale IP address URL ( http://58.093.29.223:8123, for example).

The actual bug surfaces when I physically change locations.  I've chosen the Home server, and then I leave the house.  Once that happens and my iPhone disconnects from my home network, the Home server is unreachable via its server URL.  However, the iOS companion app doesn't seem to like this.  As soon as I launch the companion iOS app outside of my home network, it displays a completely blank screen.  Then I have to pull to refresh, at which point the "You're disconnected" screen pops up.  Here, there is the server selector pop-up button at the top of the screen, **but it is not tappable**.  If I try to tap on the pop-up button, nothing happens.  It just does not register a tap, so I cannot change servers.  All of the other buttons on the "You're disconnected" screen are tappable -- the "Retry" button and the "Open App Settings" buttons at the bottom work as expected.  Just the pop-up button does not work.  After a lot of waiting and a lot of pull to refreshing, the pop-up button does eventually work, but I'm unsure why or what causes it to suddenly work.

**To Reproduce**

1. Set up two Home Assistant servers in the iOS companion app, with different URLs.  You must be able to make the URLs unreachable somehow even when you have an internet connection.
2. Connect to one of your Home Assistant servers in the iOS companion app.  Make sure the Home Assistant UI loads.
3. Close the companion app and force quit it.
4. Change internet connections so the Home Assistant server you are connected to becomes unreachable.
5. Launch the companion app.  Observe that you get a completely blank screen, but can pull to refresh.
6. Once the "You're disconnected" screen pops up, try tapping on the server selector pop-up button at the top of the screen.  Observe that nothing happens when you try to tap on it.
7. Try pulling to refresh, try tapping on the pop-up button a few more times, try opening settings, etc.  Eventually the pop-up button *will* work, but typically that happens only after a few minutes.

**Expected behavior**
I guess there are multiple issues here, but the biggest issue is that the server selector pop-up button should work immediately when it is displayed on screen.  There should be **no reason** that I can tap on the button but it doesn't show the menu to change servers.

(The other main issue is: the pop-up button to change servers should be immediately accessible at all times in the companion iOS app.  I shouldn't have to wait for the app to try to reconnect, I shouldn't have to wait for it to realize that it's disconnected, and I should be able to access it *even if* another Home Assistant server is loaded.  The pop-up button should always be accessible, and from what I can tell, it isn't.)

**Screenshots**
Here's a video showing the problem from start to finish.  It takes an entire **two minutes** for the server selection pop-up button to actually function correctly.  You can't see the interactions that I'm doing in this video, but rest assured that I am tapping that server selection pop-up button at the top of the screen many many times throughout this entire video, and it doesn't respond at all.  When you finally see the server select pop-up button show its menu in the recording (around the 2 minute mark), that's the first time it has responded at all despite dozens of taps on it.

https://github.com/user-attachments/assets/df01c3b3-d62f-439e-9ae5-bfcc26a94b6d

**Additional context**
This issue does *not* happen when visiting the Home Assistant UI directly in Safari, because I just go directly to the URL I'm interested in.  The whole "you're disconnected" screen and server select pop-up button do not exist when using stock Safari, so this is an issue with the iOS companion app specifically.  My case might be unusual, because in most instances people will probably be subscribing to Home Assistant Cloud so they probably won't have multiple servers set up, with one or the other being unreachable at all times.  So probably most users won't run into this issue.

I have read the following issue about supported remote connections in the iOS companion app: #3575 .  I believe that my configuration here **is supported**, as Tailscale is a VPN.  There are no additional location or security access restrictions, only that the Tailscale URL is reachable only when connected to the Tailscale VPN, and my home network URL is only reachable when connected to my home network.

Also, if you don't force quit the app in step 3 of the steps to reproduce, you'll see the previous Home Assistant interface but a banner will pop up at the bottom saying the app is trying to reconnect.  However, there is still no ability to change the selected server, because the server selection pop-up is shown nowhere once the Home Assistant UI has loaded.

Finally, can someone explain why there are two spinny progress indicators (the main one at the top and the smaller one at the top-right) and two places to open the same Settings menu (the gear button that sometimes shows up at the top-left and the "Open App Settings" text button at the very bottom)?  The server select pop-up button itself seems to move and reposition itself once it's finally usable, too.  It seems like this screen is a bit of a hodgepodge in design and implementation.

fengxguo:
## AI Policy

## Summary
A full-screen gestures overlay ate single-finger taps, and reconnect loader cycles hid the empty-state header for ~2 minutes (10+30+60s backoff). Hits now pass through except multi-touch/edge pans; empty state stays mounted while the web view resets.

Fixes #5198

**Test plan:** two servers, launch with the current URL unreachable, tap the server picker immediately on the disconnected screen.

## Screenshots
N/A — logic/behavior change; please verify on device as noted in the test plan.

## Any other notes

Test plan is in the Summary. CLA is signed on this account.

### DIFF
--- a/Sources/App/Frontend/WebView/HomeAssistantView/HomeAssistantViewModel.swift
+++ b/Sources/App/Frontend/WebView/HomeAssistantView/HomeAssistantViewModel.swift
@@ -38,7 +38,6 @@ final class HomeAssistantViewModel: ObservableObject {
     private let onWebViewController: ((WebViewController) -> Void)?
 
     // The standby loader remains mounted until both the minimum duration has elapsed and the frontend reconnects.
-    // Empty-state content waits for the same minimum duration so transient reload failures don't flash immediately.
     private var loaderCycleID = UUID()
     private var loaderMinimumDurationTask: Task<Void, Never>?
 
@@ -96,9 +95,10 @@ final class HomeAssistantViewModel: ObservableObject {
     }
 
     var displayedEmptyState: WebFrontendOverlayState.EmptyStateContent? {
-        guard let emptyState = overlayState.emptyState else { return nil }
-        guard isFullScreenLoaderMounted else { return emptyState }
-        return loaderMinimumDurationElapsed ? emptyState : nil
+        // Keep a published empty state on screen even while a loader cycle remounts the webview
+        // (automatic reconnect). Hiding it for the loader minimum duration was replacing the
+        // interactive header with the loader and leaving the server picker untappable.
+        overlayState.emptyState
     }
 
     var standByOpacity: Double {
@@ -135,7 +135,9 @@ final class HomeAssistantViewModel: ObservableObject {
     }
 
     func resetWebFrontend() {
-        overlayState.emptyState = nil
+        // Leave `emptyState` in place so the disconnected header (and its server picker) stays
+        // interactive while the webview is torn down and rebuilt. Successful connection still
+        // clears it via `hideEmptyState`; an explicit Retry already nils it beforehand.
         overlayState.showsNoActiveURL = false
         webViewController = nil
         beginFullScreenLoaderCycle()
@@ -153,6 +155,11 @@ final class HomeAssistantViewModel: ObservableObject {
     func handleWebViewController(_ controller: WebViewController) {
         webViewController = controller
         onWebViewController?(controller)
+        // Empty-state action closures captured the previous controller. Rebuild them on the new
+        // one so Retry / Settings / re-auth keep working across reconnect resets.
+        if overlayState.emptyState != nil {
+            controller.showEmptyState()
+        }
     }
 
     func handleWebViewLoaded(_ controller: WebViewController) {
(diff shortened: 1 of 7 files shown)


fengxguo:
Closing — not pursuing this contribution.

bgoncal (changes requested):
https://github.com/home-assistant/iOS/pull/5493#issuecomment-5388199046

fengxguo:
Closing — not pursuing this contribution.

PR #5615, opened 2026-09-02
Align widget tap behaviors with the frontend

bgoncal:
## AI Policy

## Summary

The widget customization screen offered "Toggle" to every entity, but only some domains can toggle, and "Default" didn't say what it does. This aligns the tap behaviors with the frontend's tile card:

- Defaults follow the frontend: the tile opens more info, and the icon toggles for the frontend's toggle domains (plus button, input button and scene) or opens more info otherwise. "Default" now reads "Default (Toggle)" or "Default (More info)".
- "Toggle" is only offered to the domains the frontend can toggle, honoring `supported_features` where the frontend does, and runs the same way: the entity's state picks the on or off service, so a locked lock unlocks and a group goes through `homeassistant`.
- Domain-specific behaviors are offered where they apply: Lock and Unlock, Open and Close, Turn on and Turn off for domains with an on/off pair, and the domain's main action where it isn't the toggle: Press for a button, Activate for a scene, Run for a script, Trigger for an automation.
- "Nothing" is no longer offered; more info is the fallback. An icon that only opens more info is drawn without a background.
- "Require confirmation" applies to either half of the tile, except a tap that only opens more info.

## Screenshots

The customization screen is covered by light and dark snapshot tests in `Tests/App/Settings/MagicItemCustomizationView.test.swift`.

## Any other notes

Snapshot references whose rendering changed were removed, along with new snapshot tests, so they need to be recorded on a Mac before the test job passes.


bradleyseanf:
@bgoncal


Hi, I've noticed this is now the case as of the new update today, actually I came on here to submit an issue about it but it seems this is now a new feature?

I'm not sure the UX is really that great on iOS for this, while I understand the need to align with the frontend, the icon is much smaller on iOS and is more prone to accidential clicks.

On my Homescreen on iOS I have this widget where I turn volume up and turn volume down, I need to click the button several times repeatedly (as you would on a TV remote) to turn the volume down.


image


I think it's fine on the frontend because you can hold the icon down and do different things like change color, change volume, percentage etc. 

But on iOS it's not the same so I think it's fair to not connect these two in the same way.

bgoncal:
@bradleyseanf You are still able to configure your Custom widget to do as such, it will just not be the default behavior, but the tap is configurable, you can even set it to open YouTube when tapped if desired, now, to keep consistency with the concept of tile  the icon executes the main action, tapping elsewhere opens a more-info dialog for controls

bradleyseanf:
Ah! I see you are right! Yeah that makes perfect sense then 😀

bgoncal:
This PR also makes it more clear what "Default" means for each tile tap as well as makes available more intuitive actions domain-based. Keep an eye on the next TestFlight build

PR #5712, opened 2026-09-09
Check the focus status when pushing sensors

### LINKED ISSUE #5711: Focus binary sensor stays on for hours after the Focus ends since 2026.9.0

Model Name: iPhone 17 Pro Max (iPhone18,2); also reproduced on a second phone, iPhone 17 Pro (iPhone18,1)
Software Version: iOS 26.6.1 (both phones)
App version: 2026.9.0 (2026.2874) (both phones)

2026.9.1

**Describe the bug**

Since updating to 2026.9.0, `binary_sensor.<device>_focus` sometimes stays `on` for the whole day after the Sleep Focus has ended, and no later sensor update corrects it — not background fetches, not location-triggered updates, not even launching the app.

Two occurrences in one week, on two different phones:

- Phone A: sensor `on` from 2026-09-08 00:29 until 19:02 local (18.6 h). In that window HA received 55 sensor update batches from the phone (Background Fetch, Signaled, Significant Location Change, Geographic Region, App Shortcut, Watch Context, and three `Launch` updates at 11:15, 11:32 and 17:41). Every one of them reported the Focus sensor as `on`. The phone was not in any Focus during the day.
- Phone B: sensor `on` from 2026-09-06 22:08 until 2026-09-08 00:03 (26 h), through roughly 40 update batches on 09-07, again including three `Launch` updates.

Before the update, the same two phones had flipped the sensor `on` at bedtime and `off` between 07:00 and 09:30 every single day of the previous two weeks, never longer than about 11 h `on`.

Both Focuses are scheduled Sleep Focuses whose status is shared with the app (the sensor does go `on` when they start, so the push at Focus start arrives). No Focus Filter is configured, so `sensor.focus_name` does not exist on these phones.

**To Reproduce**

1. Have a scheduled Sleep Focus with its status shared with the Home Assistant app; Focus sensor enabled.
2. Let the Focus start (sensor goes `on`) and end in the morning with the app not in the foreground.
3. On some mornings the `off` never arrives: the sensor stays `on` through every subsequent update, including opening the app.

It is not deterministic — it depends on iOS not delivering the "Focus ended" status to the app. On our phones it happened on 2 of 7 nights.

**Expected behavior**

The sensor goes `off` no later than the next sensor update after the Focus ends, as it did on the previous builds.

**Additional context**

I believe this is a side effect of #5471 (shipped in 2026.9.0). Before it, `FocusSensor` read `Current.focusStatus.status()`, which in the app process asks `INFocusStatusCenter.default.focusStatus` live on every sensor update, so a missed "ended" push was corrected at the next background fetch. Since #5471, `FocusSensor.sensors()` reads `FocusReport.current()`, and at `release/2026.9.0` that resolves the binary as

```swift
isFocused = receivedStatus?.isFocused ?? liveIsFocused()
```

i.e. the last status iOS pushed wins, and iOS is only asked live when nothing was ever pushed. Combined with what @bgoncal noted in #5467 ("iOS only wakes the app when a focus starts, never when it ends, so realtime is not guaranteed in case you turn off all focuses"), a lost "ended" push now sticks until the next push iOS happens to deliver — in our case the next evening's Focus start. The #5471 notes say "The `focus` binary sensor is unchanged", but its read path did change from live to cached.

Possible fix: a pushed status that says a Focus is running proves that this Focus's status is shared with the app, so the live `INFocusStatusCenter` query is trustworthy for it. When the stored status is `true` and the live query in the app process answers `false`, treat the Focus as ended (this keeps the #5592 behaviour for Focuses iOS cannot see, where the stored status would never have been `true`). Checking at least on app launch / foreground would already have cleared both occurrences above.

Happy to provide the full sensor history export for both phones.

g3rv4:
## AI Policy

## Summary

Since #5471 the `focus` binary sensor reports the last Focus status iOS pushed to the app and only asks `INFocusStatusCenter` when nothing was ever pushed. iOS wakes the app when a Focus starts but not reliably when one ends, so when the "ended" push does not arrive the sensor stays `on` through every later sensor update, including app launches — on two phones here it sat `on` for 18 h and 26 h after a Sleep Focus ended (#5711).

A pushed "running" only ever describes a Focus whose status the user shares with the app, and for such a Focus the live query answers truthfully. So `FocusReport` now asks iOS on every report and, when the stored status says a Focus is running but iOS says nothing is, treats the Focus as ended — provided the stored status is older than the existing switch grace period, so the tail of a Focus switch is still read as a switch. Everything else is unchanged: a pushed `false` still wins for Focuses iOS cannot see, the Focus Filter path still stands on its own, and in app extensions the "live" query is the received status itself, so nothing changes there.

Fixes #5711

## Screenshots

No UI changes.

Documentation: home-assistant/companion.home-assistant# N/A

## Any other notes

The live query is now made on every report rather than only as a fallback; it is a local `INFocusStatusCenter` read. The Focus Filter branch deliberately keeps its current behaviour (#5592): a named filter run that nothing ended stays `on` even if the live query says otherwise, because that branch also covers Focuses whose status is not shared.


g3rv4:
I suspect what failed was a flaky test, and merged `main` into the branch hoping https://github.com/home-assistant/iOS/commit/efb087b245417e266b79cc7b701d3db17ea2fd05 fixes it

Sources/Shared/Environment/FocusReport.swift
```diff
@@ -41,14 +42,25 @@ public struct FocusReport: Equatable {
         // Falling back to `name` covers state persisted before `lastKnownName` existed.
         let lastKnownName = filterState?.lastKnownName ?? filterState?.name
 
+        let liveStatus = liveIsFocused()
+
         let isFocused: Bool?
         if let filterState, filterState.name?.isEmpty == false,
```
bgoncal:
```suggestion
        if let filterState, !filterState.name?.isEmpty,
```

Sources/Shared/Environment/FocusReport.swift
```diff
@@ -41,14 +42,25 @@ public struct FocusReport: Equatable {
...
            !hasEnded(filterState: filterState, receivedStatus: receivedStatus) {
             // A filter only runs with a name when a Focus starts — the nil-name run iOS makes on
             // deactivation must not count — and nothing has told us it ended since.
             isFocused = true
+        } else if let receivedStatus, receivedStatus.isFocused == true, liveStatus == false,
```
bgoncal:
```suggestion
        } else if let receivedStatus, receivedStatus.isFocused, !liveStatus,
```

Sources/Shared/Environment/FocusReport.swift
```diff
@@ -41,14 +42,25 @@ public struct FocusReport: Equatable {
...
+                  Current.date().timeIntervalSince(receivedStatus.date) > switchGracePeriod {
+            // iOS wakes us when a Focus starts but not reliably when one ends, so the "ended" push
+            // can simply never arrive and the last push would stand until the next Focus starts.
+            // A pushed "running" only ever describes a Focus whose status the user shares, which
+            // is the one case where asking iOS directly answers truthfully — so once that push has
+            // had the switch window to settle, a live "nothing is running" is the ending we were
+            // never told about.
+            isFocused = false
```
bgoncal: If the goal is "Always rely on iOS focus state (not focus name) live information" you can comment that directly, it makes sense alone without the rest of the explanation, it is also the more correct approach indeed, if we ask iOS "Hey is the focus in use?", whatever it replies should be the source of truth, if it replies incorrectly then it's an iOS bug

Tests/Shared/Sensors/FocusSensor.test.swift
```diff
@@ -140,6 +144,28 @@ class FocusSensorTests: XCTestCase {
         XCTAssertEqual(focusSensor.State as? Bool, false)
     }
 
+    /// The bug behind #5711: iOS wakes us when a Focus starts but not reliably when one ends, so
+    /// a pushed "running" whose "ended" push never came stood until the next Focus started. iOS
+    /// only pushes "running" for a Focus whose status the user shares, so the live status is
+    /// truthful for it — but inside the switch window it can still describe the Focus that just
+    /// ended, so the push has to be older than that window before a live "not focused" ends it.
+    func testStaleReceivedRunningEndsOnceTheLiveStatusSaysNotFocused() throws {
```
bgoncal, line 152: Not sure how I feel about documenting fixes in test headers, but let's keep it for now, something for me to reflect on

bgoncal (changes requested):
Thank you for looking into this, just minor comments, also, were you able to test this solution for a reasonable period of time? Focus reporting has presented itself a bit unreliable in iOS reporting system

g3rv4:
> Thank you for looking into this, just minor comments, also, were you able to test this solution for a reasonable period of time? Focus reporting has presented itself a bit unreliable in iOS reporting system

I haven't tested it. I don't have an Apple dev account, and my personal account can't build it because the app uses capabilities that aren't available for personal accounts:

image


I trust your experience with the iOS focus reporting limitations, I'm happy to close this PR and rely on an shortcut automation to keep tabs of my focus state if you'd prefer to go that route.

bgoncal:
The PR is not bad, I was jus curious if you were able to confirm the implementation, check our readme file on how to setup your xcode for your account, you can also remove some features that require entitlement and then it will build to your own device, I will test it myself of course in parallel.

g3rv4:
> The PR is not bad, I was jus curious if you were able to confirm the implementation, check our readme file on how to setup your xcode for your account, you can also remove some features that require entitlement and then it will build to your own device, I will test it myself of course in parallel.

I can remove the entitlements, but from [here](https://developer.apple.com/documentation/UserNotifications/handling-communication-notifications-and-focus-status-updates), it seems like I need the Communication Notifications capability, which I can't sign for.

Would you be able to do a testflight build? I'd be happy to run that on both devices to validate it works as expected.

bgoncal (approved):

PR #5764, opened 2026-09-13
Fix WatchWidgets bundle identifier to use BUNDLE_ID_PREFIX

alexspeller:
## AI Policy

## Summary

`Extensions-WatchWidgets` sets its `PRODUCT_BUNDLE_IDENTIFIER` to a hardcoded `io.robbie.HomeAssistant[.dev].watchkitapp.WatchWidgets` in both the Debug and Release configurations, instead of deriving it from `$(BUNDLE_ID_PREFIX)` the way every other target does.

This breaks device builds for contributors who set a custom `DEVELOPMENT_TEAM` / `BUNDLE_ID_PREFIX` via `Configuration/HomeAssistant.overrides.xcconfig` (the flow documented in the README): automatic signing fails because the `io.robbie.…` WatchWidgets App ID cannot be registered under a team other than the project's own.

The fix points both configurations at `$(BUNDLE_ID_PREFIX).HomeAssistant$(BUNDLE_ID_SUFFIX).watchkitapp.WatchWidgets`, matching the WatchApp target (which uses `PROVISIONING_SUFFIX = .watchkitapp`) and the rest of the project. With the default `io.robbie` prefix the resulting bundle identifier is unchanged.

## Screenshots

N/A — build configuration change, not user-facing.


bgoncal (approved):
Great catch! Thank you

PR #5773, opened 2026-09-14
Extract shared webhook secretbox crypto

gbyo:
## AI Policy

* [x] I have read the [[Open Home Foundation AI Policy](https://developers.home-assistant.io/docs/ai_policy)](https://developers.home-assistant.io/docs/ai_policy).

* [ ] I have not used AI for this contribution.
* [x] AI assistance was used for this contribution.
* [ ] AI fully generated the code for this contribution, but I've reviewed and understood it before submitting and will respond without AI during review.

## Summary

Moves the existing webhook secretbox encryption code into a small local WebhookCrypto Swift package.

WebhookPayloadCrypto still handles the existing iOS/watchOS webhook path, but the actual secretbox implementation now lives in its own module.

Needed for the iOS 27 Remote Now Playing extension so it can use the same encrypted webhook transport without depending on the full Shared module.

## Screenshots

n/a

Documentation: Not applicable

## Any other notes

Split from #5690


Sources/WebhookCrypto/WebhookSecretBox.swift
```diff
@@ -0,0 +1,52 @@
+import Foundation
+import Sodium
+
+/// The secretbox primitive Home Assistant's `mobile_app` webhook wraps payloads in.
+///
+/// Kept in a low-level source set so lightweight targets can compile it directly later without
+/// linking the full `Shared` module. The phone's webhook path and the watch client use this one
+/// implementation rather than diverging copies. It depends on nothing but Foundation and `Sodium`,
+/// which is statically linked and contributes only the secretbox and base64 objects.
+enum WebhookSecretBox {
```
bgoncal: Can this be a local SPM package instead of a plain folder? Every other top level dir in `Sources/` is either an Xcode target or a package (HANetworking, HAModels, HAUtilities, HAIconic, SharedPush).

As a folder there is no enforced boundary, it compiles into `Shared` and can reach Shared internals, and when the extension adds the same synced group it compiles a second copy, so you end up with two unrelated `WebhookSecretBox` types instead of one shared module.

A small package depending on swift-sodium pinned `exact: "0.11.0"` follows what HANetworking already documents.

HomeAssistant.xcodeproj/project.pbxproj
```diff
@@ -2534,6 +2534,7 @@
 		42F467200000000000000001 /* NotificationSender */ = {isa = PBXFileSystemSynchronizedRootGroup; explicitFileTypes = {}; explicitFolders = (); path = NotificationSender; sourceTree = "<group>"; };
 		42FD0000009000510051FC9B /* VoiceToolsServer */ = {isa = PBXFileSystemSynchronizedRootGroup; explicitFileTypes = {}; explicitFolders = (); path = VoiceToolsServer; sourceTree = "<group>"; };
 		42FD0000009100510051FC9B /* VoiceToolsServer */ = {isa = PBXFileSystemSynchronizedRootGroup; explicitFileTypes = {}; explicitFolders = (); path = VoiceToolsServer; sourceTree = "<group>"; };
+		42FE0004300000000000RM04 /* WebhookCrypto */ = {isa = PBXFileSystemSynchronizedRootGroup; explicitFileTypes = {}; explicitFolders = (); path = WebhookCrypto; sourceTree = "<group>"; };
```
bgoncal: `42FE0004300000000000RM04` is the only non hex object ID in the whole project file. Xcode parses it, but it breaks the convention and Xcode may rewrite it on save and add noise to a future diff. Could you make it hex?

Sources/WebhookCrypto/WebhookSecretBox.swift
```diff
@@ -0,0 +1,52 @@
...
+    enum CryptoError: Error, Equatable {
+        case encode
+        case seal
+        case base64
+        case decode
+        case open
+    }
+
+    /// `object` (a JSON value) sealed with `secret` and base64-encoded, ready to be sent as a
```
bgoncal: This says "a JSON value" but the original said "a JSON object", and that difference matters: `JSONSerialization.data(withJSONObject:)` without `.fragmentsAllowed` raises an ObjC exception for a top level fragment, so it crashes instead of throwing.

No bug today since every caller passes a dictionary. Either keep "JSON object" or guard with `isValidJSONObject` and throw `.encode`.

Sources/WebhookCrypto/WebhookSecretBox.swift
```diff
@@ -0,0 +1,52 @@
+import Foundation
+import Sodium
+
+/// The secretbox primitive Home Assistant's `mobile_app` webhook wraps payloads in.
+///
+/// Kept in a low-level source set so lightweight targets can compile it directly later without
+/// linking the full `Shared` module. The phone's webhook path and the watch client use this one
```
bgoncal: Two things in this comment: the watch and phone already shared one implementation before this PR, so that sentence describes the status quo, and "contributes only the secretbox and base64 objects" is a linker outcome that depends on dead stripping. `Sodium()` is a struct with 19 stored sub structs.

Worth measuring against the 6mb budget rather than stating it here, since the comment will rot.

Sources/Shared/API/Webhook/Networking/WebhookPayloadCrypto.swift
```diff
@@ -13,24 +17,15 @@ enum WebhookPayloadCrypto {
     /// `data` (a JSON object) sealed with `secret` and base64-encoded, ready to be sent as the
     /// request's `encrypted_data` value.
     static func encrypt(_ data: Any, secret: [UInt8], sodium: Sodium = Sodium()) throws -> String {
-        let jsonData = try JSONSerialization.data(withJSONObject: data, options: [.sortedKeys])
-
-        guard let jsonStr = String(data: jsonData, encoding: .utf8) else {
+        do {
+            return try WebhookSecretBox.seal(data, secret: secret, sodium: sodium)
+        } catch WebhookSecretBox.CryptoError.encode {
```
bgoncal: `WebhookPayloadCrypto.CryptoError` is never caught or asserted anywhere in Sources or Tests. Its only consumer is the log line in `WebhookRequest.swift:64`, and both enums interpolate the same text, so this mapping produces identical output.

Could we drop it and the duplicate enum, and keep only the `decrypt` mapping, which is asserted by the tests? These three throws are also new untestable lines, so they will hurt patch coverage.

Sources/Shared/API/Webhook/Networking/WebhookPayloadCrypto.swift
```diff
@@ -13,24 +17,15 @@ enum WebhookPayloadCrypto {
...
-        guard let encryptedData: Bytes = sodium.secretBox.seal(
-            message: jsonStr.bytes,
-            secretKey: .init(secret)
-        ) else {
+        } catch WebhookSecretBox.CryptoError.seal {
             throw CryptoError.seal
-        }
-
-        guard let b64payload = sodium.utils.bin2base64(encryptedData, variant: .ORIGINAL) else {
+        } catch WebhookSecretBox.CryptoError.base64 {
```
bgoncal: If the mapping stays: these catches are not exhaustive, so a new case in `WebhookSecretBox.CryptoError` would leak past the wrapper with no compiler error. `catch let error as WebhookSecretBox.CryptoError { switch error { ... } }` would turn that into a build failure.

bgoncal (changes requested):
Thanks for splitting this out, the extraction itself is clean and I checked the behaviour is identical (sortedKeys, ORIGINAL variant, the empty payload `()` case, error mapping 1:1). Target membership matches the sibling synced groups and it merges cleanly with main.

Main thing I would like changed is the new `Sources/WebhookCrypto` folder, I think it should be a local SPM package like the rest of our low level code. The others are small.

Sources/WebhookCrypto/WebhookSecretBox.swift
gbyo: That makes sense, and it would give this an actual module boundary instead of just relying on the folder layout.

I’ll switch it over.

HomeAssistant.xcodeproj/project.pbxproj
gbyo: Fixed this as part of the spm change, the new project IDs are hex now.

PR #5824, opened 2026-09-16
Live Activity: add when_start as the explicit timer start

DerZade:
## AI Policy

## Summary
This PR adds a new `when_start` field for (live) notifications, which can be used for anchoring chronometers.

Currently there is no way to update chronometers (let them run longer / shorter) without them starting to count from the start again. This is true for both counting up and counting down. On iOS this was even more visible, because there a progress bar is displayed with the counter. So if for example my washer re-estimates the remaining time in the middle of the cycle, the progress bar resets to 100% and "counts down" from the new remaining time. 

The only thing which I found was remotely touching on this was [`carryForwardChronometerAnchor` in the iOS App](https://github.com/home-assistant/iOS/blob/edc7153a5e99f637a27a69e8285887114456c985/Sources/Shared/LiveActivity/LiveActivityRegistry.swift#L431-L447), which only works for count-up chronometers if the same relative duration is passed. 

With `when_start` the user has full control over the starting anchor for both count-down and count-up chronometers. On Android it applies to count-ups sent as a negative relative `when`. Count-downs are unaffected, since they are just text and an absolute `when` is already a stable anchor.

`when_start` is always a Unix timestamp. `when_relative` applies to `when` only. Payloads without `when_start` keep the existing carry-forward behavior in iOS unchanged. 

Documentation: home-assistant/companion.home-assistant#1427

## Any other notes

PR for Android: home-assistant/android#7489


bgoncal:
I'm concerned about the complexity of how the payload is generated now, perhaps the original decisions on how to render the chronometer live activity weren't user-friendly enough. We were having a chat about this in here if you want to jump in https://discord.com/channels/330944238910963714/1346948551892009101/1550072534056177695

DerZade:
> I'm concerned about the complexity of how the payload is generated now, perhaps the original decisions on how to render the chronometer live activity weren't user-friendly enough. We were having a chat about this in here if you want to jump in https://discord.com/channels/330944238910963714/1346948551892009101/1550072534056177695

First up sorry for the delayed response.

I totally get what you mean with complexity. I thought quite a bit on how to "design the API" before actually deciding on this solution. I still think tho that within the current constraints, and how the rest of the fields look, it is the cleanest and most consistent.

In my option the `when_start` field isn't really adding a lot of complexity. It is optional and can only be a timestamp. The issue for me is more that the `when` field already tried to solve too many things at once, by allowing relative and absolute values as well as negative and positive values for count up / count down, that probably should have been solved with two or three fields from the get go. That is imho what makes the API so "complex". 

I think if you would have just had two fields from the get go, it would have been more elegant. The `when` currently tries to be a point in time (an anchor) and a duration at the same time. Splitting these up into two fields would have made my PR obsolete and would have made the API more intuitive:

```yaml
data:
  chronometer_anchor: {{ now() }} # maybe now is default value
  chronometer_duration: 30 # in seconds; count up from anchor
```

```yaml
data:
  chronometer_anchor: {{ as_timestmap(now()) + 30 }} 
  chronometer_duration: -30 # in seconds; count down to anchor
```

Yes that would have meant a bit of math for some users, but for anyone who doesn't want to fiddle around with dates and just wants a simple count up / count down. It would have been more or less the same (just needing to set `chronometer_duration`)

---

But sadly we now already have `when` as our status quo, so `when_start` was the best thing I could come up with :P  

@bgoncal  What would be your suggestion on how to solve this?

PR #5841, opened 2026-09-18
fix: resolve synced shortcut server references on single-server devices

### LINKED ISSUE #5504: "Perform action" shortcut fails with "No server provided" on second device after iCloud sync, despite Server field showing "Home"
Model Name: iPhone 16
Software Version: 26.6
App version: 2026.7.5 (2026.2744)

2026.8.3

**Describe the bug**
A Shortcut containing a Home Assistant "Perform action" step, created on one device (macOS) and synced to another (iPhone) via iCloud, fails on the second device with a "No server provided" error - even though every field in the action is populated correctly: Server displays "Home", Action displays "Turn on fan", and the action data is valid (`{"entity_id": "fan.probreeze"}`).

The failure is a contradiction between the displayed state and the resolved state. The Server field shows "Home" as the selected server, yet at runtime the action reports that no server is provided. Tapping "Show" on the error opens the shortcut editor with a red "No server provided" banner at the top while the Server row still reads "Home" (see screenshots).

The underlying server reference appears to be per-installation: the shortcut and its "Home" display label sync via iCloud, but the reference does not resolve on a device other than the one where it was selected.

Attempting to fix it on the second device creates an unresolvable loop: re-selecting the server on the iPhone clears the Action selection (Action appears to be scoped to the server), requiring the Action to be re-selected; and once the shortcut is corrected on the iPhone, the change syncs back via iCloud and the original macOS copy then fails in the same way. A single shortcut therefore cannot be made to work across both devices simultaneously.

**To Reproduce**
1. On macOS, create a Shortcut with a Home Assistant "Perform action" step. Set Server = Home, Action = "Turn on fan" (`fan.turn_on` — any action reproduces it), action data = `{"entity_id": "fan.probreeze"}`. Confirm it runs correctly on macOS.
2. Allow the Shortcut to sync to an iPhone via iCloud (same Apple ID; companion app installed and signed in to the same single server on both devices).
3. Run the Shortcut on the iPhone. → Fails with "No server provided", though the Server field shows "Home".
4. Tap "Show" on the error. → The editor opens with a red "No server provided" banner while the Server row still displays "Home".
5. On the iPhone, tap the Server row and re-select Home. → The Action selection is cleared and must be re-selected.
6. Once corrected on the iPhone, the change syncs back to macOS via iCloud. → The macOS copy now fails with the same error.

**Expected behavior**
A Shortcut created on one device and synced to another should resolve the selected server on any device where that server is registered and signed in - particularly in the single-server case, where resolution is unambiguous.

Failing that, the UI should not display "Home" as the selected server when that reference does not resolve on the current device. The displayed selection and the resolver disagree, which makes the action appear correctly configured while it cannot run — the field should show no selection (prompting re-selection) rather than a stale label that masks the unresolved reference.

**Screenshots**
Image
Image
Image

**Additional context**
* Second device is macOS: MacBook Air M2 (2022), macOS 26.5.1 (25F80), companion app version 2026.7.4. Note the app versions differ slightly between devices (iOS 2026.7.5 / macOS 2026.7.4).
* Both devices are signed in to the same single Home Assistant server ("Home") in the companion app.
* In some other instances the same underlying failure surfaced as a generic "unknown error occurred" dialog rather than the explicit "No server provided" text; the explicit form is shown in the attached screenshots.
* Happy to provide companion app logs on request.

mvanhorn:
## AI Policy

## Summary

In IntentServerAppEntitiy.swift, centralize entity lookup through the existing getServer(): prefer an exact local ID match, otherwise resolve a nonempty unknown ID only when exactly one local server exists and mirror restoration is not pending. Update entities(for:) to reconstruct requested entities through that same resolver, retaining each requested ID rather than substituting the local server's ID; getInfo() will then provide the locally resolved display name while suggested entities continue to use local IDs. In IntentActionEntityQuery.actionEntities(), fetch definitions from the resolved local server but construct action entities with the selected intent.server.id, so lookup, suggestions, response-support metadata, and persisted action identifiers agree with the retained server reference.

A Perform action shortcut created on macOS retains its server label after iCloud sync but fails with “No server provided” on an iPhone configured for the same Home Assistant instance. Reselecting the server changes the shortcut's server reference and clears its dependent action, causing the original device to fail after sync. The local code resolves IntentServerAppEntity IDs strictly against Current.servers, while IntentActionEntity persists serverId::domain.service and PerformActionAppIntent checks that action.serverId equals server.id before resolving the server. The report supplies a concrete two-device reproduction with one configured server on each device; no logs or hardware-specific diagnosis are needed to test the underlying lookup failure.

Fixes #5504

## Screenshots

No user-visible surface changes in this PR, so there is nothing to show.

## Any other notes

Nothing beyond what is described above.

AI was used for assistance.


bgoncal (changes requested):
Thank you for your PR, since you marked AI fully generate the code, please attach a video showing the old version of the app, then creating the shortcut, then upgrade the app (to the code in main branch) where it would fail, finally show it upgrading again to this branch code and succeeding.

Mark as ready for review afterwards

bgoncal:
I was playing with this and I have a different fix, a bit simpler, can you check? https://github.com/home-assistant/iOS/pull/5842

mvanhorn:
Checked it, and #5842 handles the case I was targeting. I think it's the better shape, so I'm happy to close mine in favor of it.

The core conclusion is the same one I reached: the stored serverId on a synced shortcut can be stale or point at a server the device no longer has, and on a single-server install the right move is to fall back to the only server rather than throw noServer. Putting that behind `IntentServerAppEntity.shortcutServer(for:)` is cleaner than what I did, since the intents share one helper instead of repeating resolution logic at each call site.

One small thing worth a second look, not a blocker. In GetCameraSnapshotAppIntent and PerformActionAppIntent the ownership guard changed from

```swift
guard camera.serverId == server.id
```

to

```swift
IntentServerAppEntity.shortcutServer(for: camera.serverId)?.identifier == server.identifier
```

On a multi-server device that fails closed the way you'd want: an unresolvable id returns nil and the guard rejects. On a single-server device the fallback returns the only server for any unresolvable id, so the comparison is always true and the "does this entity belong to this server" check stops doing anything. Harmless today, because there is only one server it could belong to, but it is the spot that would quietly stop guarding if the fallback is ever widened. Might be worth a short comment there noting the check is intentionally vacuous in the single-server case.

Mine is still a draft, so no need to review it. I'll close it once yours lands.

bgoncal:
Thanks for checking, I merged the other PR, I'll be closing this one, feel free to reopen if needed

